PR #1, opened 2022-06-27
resolve pathing issues when metric-gardener is installed via npm

Chrisp485:
At the start of the app, it now will check if metric-gardener is installed locally as a dependecy or global on your system and adjust the path accordingly for the config file. Next step would be to publish now to npm and see if what was tested locally works aswell if you download metric-gardener via npm. 

### To test this PR:
Local  installation In any npm project root: 
```

$ npm install path/to/mg/onyoursystem
$ npm exec metric-gardener -- parse input/path -o output/path.json
```
Global installation (go to root of metric gardener repo on your system:
```
$ npm install -g
$ metric-gardener parse input/path -o output/path. json

```


ce-bo (approved):
LGTM but would you please address my remarks?
Furthermore, could you please have a look in the README.md. I think the old config name and path should be changed there. And feel free to delete the old config that should not broke anything.

Issue #15, opened 2023-12-07
Setup Renovate


ce-bo:
Integrate Renovate for automated dependency updates like it is used in CodeCharta repo.

michaelalper-cezik-mw:
Renovate has successfully been set up in the metric-gardener repository.

**Important note:**
Each time tree-sitter grammar is updated it is necessary to manually run the import-script. More information on how to run the script can be found here: https://github.com/MaibornWolff/metric-gardener/blob/main/UPDATE_GRAMMARS.md

ResistantBear:
We probably close this issue once we verified that the weekend-schedule and grouping has worked as expected.

ce-bo:
@ResistantBear does it work and can this issue be closed?

ResistantBear:
Jep. Here and there, there occur some minor issues with the dependencies for which we have to re-group the dependency updates, but we have opened new issues for these.

ResistantBear:
Jep. Here and there, there occur some minor issues with the dependencies for which we have to re-group the dependency updates, but we have opened new issues for these.

PR #17, opened 2023-12-22
Updated tree-sitter, added clarification regarding the build process to Readme, added test script entry

ResistantBear:
Updated to tree-sitter 0.20.6, which is the latest version of tree-sitter at the time of this pull request, runs now with Node 20.10.0 (LTS).

I have added a test script entry in package.json that runs "jest" in order for "npm test" to work. This should be changed to the new method once #20 is implemented.

Added entry for the supported node version in package.json

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 #### Install required build tools:
 
--   This project is based upon Node.js. Therefore, there has to be a recent version of Node.js installed on your machine (Version 18 or newer).
+-   This project is based upon Node.js. Therefore, there has to be a recent version of Node.js installed on your machine (version 18 or newer).
 -   Ensure that you have installed all build tools necessary to compile the native C/C++ parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
 
 #### Install project and parse your sources:
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 #### Install required build tools:
 
-Ensure that you have installed all build tools necessary to compile the native C++ parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
+Ensure that you have installed all build tools necessary to compile the native parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
 
 #### Install project and parse your sources:
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 #### Install required build tools:
 
-Ensure that you have installed all build tools necessary to compile the native parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
+Ensure that you have installed all build tools necessary to compile the native C/C++ parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
 
 #### Install project and parse your sources:
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 #### Install required build tools:
 
--   This project is based upon Node.js. Therefore, there has to be a recent version of Node.js installed on your machine.
+-   This project is based upon Node.js. Therefore, there has to be a recent version of Node.js installed on your machine (Version 18 or newer).
 -   Ensure that you have installed all build tools necessary to compile the native C/C++ parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
 
 #### Install project and parse your sources:
--- a/README.md
+++ b/README.md
@@ -19,7 +19,8 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 #### Install required build tools:
 
-Ensure that you have installed all build tools necessary to compile the native C/C++ parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
+-   This project is based upon Node.js. Therefore, there has to be a recent version of Node.js installed on your machine.
+-   Ensure that you have installed all build tools necessary to compile the native C/C++ parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
 
 #### Install project and parse your sources:
 
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 ### Usage
 
+#### Install required build tools:
+
+Ensure that you have installed all build tools necessary to compile the native C++ parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
+
 #### Install project and parse your sources:
 
 -   `npm install`
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 #### Install required build tools:
 
--   This project is based upon Node.js. Therefore, there has to be a recent version of Node.js installed on your machine (version 18 or newer).
+-   This project is based upon Node.js. Therefore, there has to be a recent version of Node.js installed on your machine (see `engines` entry of package.json for version requirements).
 -   Ensure that you have installed all build tools necessary to compile the native C/C++ parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
 
 #### Install project and parse your sources:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -17,6 +17,11 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 ### Usage
 
+#### Install required build tools:
+
+-   This project is based upon Node.js. Therefore, there has to be a recent version of Node.js installed on your machine.
```
ce-bo, line 22: Lets specify a lower limit for the node version.

ce-bo: We should Support at least the latest LTS (v18).

package.json
```diff
@@ -6,6 +6,7 @@
     "bin": "dist/app.js",
     "scripts": {
         "start": "tsc && node dist/app.js",
+        "test": "jest",
```
ce-bo, line 9: I think some configuration is needed. You could check other jest config files e.g. in the Codecharta repo (visualization). 

Maybe Running jest without and config does not work because we are not sticking to the jest standards (test file names, default pathes for test resources, etc. ) or simply a Minimum configuraion is needed.

README.md
ResistantBear: Ok. Should I explicitely verify that the program runs with version 18?

ce-bo: No, mentioning it in the readme as a requirement should be fine.

package.json
ce-bo: AFAIK the config file has been moved to src/parser/config/nodeTypesConfig.json
For sure it was forgotten to adjust the test. 

Hmm, what is types uses for?

BridgeAR: I suggest to switch to the built-in Node.js test runner or mocha.

Jest is slow and does pretty bad things internally that features are not working as expected (e.g., `instanceof` being broken, it manipulates assertion libraries output and handling and more).

README.md
```diff
@@ -17,6 +17,11 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 ### Usage
 
+#### Install required build tools:
+
+-   This project is based upon Node.js. Therefore, there has to be a recent version of Node.js installed on your machine (version 18 or newer).
```
BridgeAR, line 22: The required Node.js version may be specified in the package.json. That way there is no further need to document the version in the readme.

https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

package.json
ResistantBear: > AFAIK the config file has been moved to src/parser/config/nodeTypesConfig.json For sure it was forgotten to adjust the test.
> 
> Hmm, what is types uses for?

The issue of the wrong path is not within the test cases, but within the application logic (src/parser/CouplingCalculator.ts). 

However, I have changed the path to the file you suggested and that fixed the issue. The test cases still fail, but this is because the generated JSON contains the absolute file paths to the parsed files on several places, which has of course a different prefix compared to `C:\Users\CedrikB\projects\node-parser"` in my case. I guess this is unavoidable as long as the absolute file path is included in the JSON and the test case compares the complete JSON output. 
As the rest of the output is identical, I guess that the application code runs fine now.

Regarding the types field, as this is always empty, I don't know what it is used for. I will look into the source code to find some hints about it.

ResistantBear: > I suggest to switch to the built-in Node.js test runner or mocha.
> 
> Jest is slow and does pretty bad things internally that features are not working as expected (e.g., `instanceof` being broken, it manipulates assertion libraries output and handling and more).

That sounds reasonable. However, as the current issue is not related to a problem with Jest, I have created the new issue #20 for this suggestion.

README.md
ResistantBear: Thanks for the hint. I have changed the package.json and readme accordingly.

package.json
```diff
@@ -20,7 +24,7 @@
     "dependencies": {
         "copyfiles": "^2.4.1",
         "javascript-lp-solver": "^0.4.24",
-        "tree-sitter": "^0.20.0",
+        "tree-sitter": "^0.20.6",
         "tree-sitter-c-sharp": "^0.19.0",
```
BridgeAR, line 28:
```suggestion
        "tree-sitter-c-sharp": "^0.20.0",
```

package.json
```diff
@@ -20,7 +24,7 @@
     "dependencies": {
         "copyfiles": "^2.4.1",
         "javascript-lp-solver": "^0.4.24",
```
BridgeAR, line 26: This seems to be unused?

@ce-bo ?

package.json
```diff
@@ -20,7 +24,7 @@
...
-        "tree-sitter": "^0.20.0",
+        "tree-sitter": "^0.20.6",
         "tree-sitter-c-sharp": "^0.19.0",
         "tree-sitter-cpp": "^0.20.0",
```
BridgeAR, line 29:
```suggestion
        "tree-sitter-cpp": "^0.20.3",
```

package.json
```diff
@@ -20,7 +24,7 @@
...
         "tree-sitter-go": "^0.19.1",
```
BridgeAR, line 30:
```suggestion
        "tree-sitter-go": "^0.20.0",
```

BridgeAR (commented):
Since this is about updating tree-sitter, I guess we could also update all other tree-sitter dependencies.

Could you please open a ticket to install renovate for automated dependency updates? :-)

ResistantBear:
I tried updating all tree-sitter dependencies to their newest versions, however, updating "tree-sitter-php" reproducibly leads to the failure of the test case "parses PHP McCabeComplexity metric => should count if statements correctly". This change also increases the deviation to the snapshot in the test case "parsing PHP dependencies".

ResistantBear:
> Since this is about updating tree-sitter, I guess we could also update all other tree-sitter dependencies.
> 
> Could you please open a ticket to install renovate for automated dependency updates? :-)

I tried updating all tree-sitter dependencies to their newest versions, however, updating "tree-sitter-php" reproducibly leads to the failure of the test case "parses PHP McCabeComplexity metric => should count if statements correctly". This change also increases the deviation to the snapshot in the test case "parsing PHP dependencies".

The renovate-thing should be already covered with #15 I guess?

package.json
ResistantBear: > Regarding the types field, as this is always empty, I don't know what it is used for. I will look into the source code to find some hints about it.

I have found out from the Git history that @ce-bo has added this "types" field in src/commands/outputMetrics.ts to store the "namespace" coupling metric in the commit https://github.com/MaibornWolff/metric-gardener/commit/e134ae4c7be9455262dee4b4719671c1a1c08134
However, this whole coupling metric and the code to add it to the field has been removed in https://github.com/MaibornWolff/metric-gardener/commit/2becab6f8d454c8984d84f45c262eaeee8f9a1d5, while the field "types" still remains.
I guess that we should remove the now unused field "types" again?

ResistantBear:
Since the project is not installable at all with current versions of node.js without this change here and there seems to be more to updating these other tree-sitter dependencies (as I understand: grammars for the respective languages) than just changing the version number, we might open a new issue for updating these other dependencies and fixing this PHP issue that occurs by doing so?

ResistantBear:
> Since the project is not installable at all with current versions of node.js without this change here and there seems to be more to updating these other tree-sitter dependencies (as I understand: grammars for the respective languages) than just changing the version number, we might open a new issue for updating these other dependencies and fixing this PHP issue that occurs by doing so?

I have opened issue #24 for updating the other tree-sitter dependencies/grammars now.

package.json
BridgeAR: I suggest to move this conversation to an issue. It seems unrelated to what should be achieved here?

Adding a general test script seems good enough, even if it's not possible to execute it as it's broken.

BridgeAR (approved):
I would update all tree-sitter ones that are not causing any issues and merge it afterwards. You could also remove one unused library, if I am not mistaken. Otherwise I would forget about it later.

package.json
ResistantBear: As it has turned out that this issue is not related to the test script, thats correct.
I have opened the issues #25 and #26 for further discussion.

ResistantBear:
> I would update all tree-sitter ones that are not causing any issues and merge it afterwards. You could also remove one unused library, if I am not mistaken. Otherwise I would forget about it later.

Thinking about these dependencies/grammars, the Readme states that there are actions to be performed in case the grammars are updated, including the manual mapping of expressions (see https://github.com/MaibornWolff/metric-gardener#maintenance-of-tree-sitter-grammars). This explains why the program fails when updating the grammar without updating these mappings. I am therefore hesitant to change any of them until clarifying with @ce-bo how to update the existing mappings efficiently. I have added the issues #24 for this.

Regarding the unused library, I have opened issue #27 to clarify that.

Issue #2, opened 2022-08-05
Stop verbose logging during parsing


BridgeAR:
Logging all entries is expensive and very verbose. This should should be behind the debug env. As such, switch to `util.debuglog()`.

PR #19, opened 2023-12-27
Replaced most console.log outputs with conditional debuglog outputs, added some documentation

ResistantBear:
Done as suggested by the issue #2. The debug outputs can be enabled by adding the environment variable `NODE_DEBUG` with the value `metric-gardener`.

When the debug-outputs are disabled, the performance of analyzing the jUnit 4 code base improves from around 10-11 seconds before this change to around 9-10 seconds. However, with debug-outputs enabled, the execution time seems to be a bit slower than before, at 11-12 seconds. This could be related to additional outputs, like the "METRIC-GARDENER" before each line introduced by using debuglog, or the re-enabled printing of the query statements.

Should probably not be merged before #17 is approved, as it builds upon it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@ One parser for all languages!
 No need to build your source code.
 This static source code parser calculates simple software quality metrics for quite a few languages (more will be added).
 
-
 An experimental approach is implemented to approximate couplings metrics for C# and PHP without you having to build the source code.
 This is quite slow and can take up to one or two hours but can provide good results.
 
@@ -84,6 +83,11 @@ For this, reimport grammars for supported languages from tree-sitter by:
 -   `npm run start -- import-grammars`
 
 This will have no effect until you have mapped the changed and new expressions manually to ./src/parser/config/nodeTypesConfig.json
+
+### Enable debug prints
+
+There are additional outputs about the metric calculation process that can be enabled by setting the
+environment variable `NODE_DEBUG` to `metric-gardener`.
 
 ### TODOs
 
--- a/src/parser/GenericParser.ts
+++ b/src/parser/GenericParser.ts
@@ -5,6 +5,9 @@ import { Configuration } from "./Configuration";
 import { MetricCalculator } from "./MetricCalculator";
 import { CouplingCalculator } from "./CouplingCalculator";
 import { CouplingResult, MetricResult } from "./metrics/Metric";
+import { debuglog } from "node:util";
+
+const dlog = debuglog("metric-gardener");
 
 export class GenericParser {
     private readonly config: Configuration;
@@ -23,7 +26,7 @@ export class GenericParser {
             []
         );
 
-        console.log(" --- " + parseFiles.length + " files detected", "\n\n");
+        dlog(" --- " + parseFiles.length + " files detected\n\n");
 
         let fileMetrics = new Map<string, Map<string, MetricResult>>();
         if (this.config.parseMetrics) {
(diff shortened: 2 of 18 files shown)


src/parser/GenericParser.ts
```diff
@@ -5,14 +5,32 @@ import { Configuration } from "./Configuration";
 import { MetricCalculator } from "./MetricCalculator";
 import { CouplingCalculator } from "./CouplingCalculator";
 import { CouplingResult, MetricResult } from "./metrics/Metric";
+import { debuglog } from "node:util";
 
+const dlog = debuglog("metric-gardener");
```
BridgeAR, line 10: This is going to return a wrapper to log things. If the logging time itself should also be reduced, it's possible to use the callback. That way the additional overhead should be lower.

https://nodejs.org/api/util.html#utildebuglogsection-callback

README.md
```diff
@@ -17,6 +17,10 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 ### Usage
 
+#### Install required build tools:
+
+Ensure that you have installed all build tools necessary to compile the native parts of the required tree-sitter package: https://github.com/nodejs/node-gyp#installation
```
BridgeAR, line 22: This is only needed in case tree-sitter has no pre-build binaries for the specific Node.js version.

BridgeAR (approved):
LGTM. Just left two nits.

src/parser/GenericParser.ts
ResistantBear: Ok, thanks for the hint. I have changed that.

src/parser/resolver/callExpressions/AbstractCollector.ts
```diff
@@ -40,7 +45,7 @@ export abstract class AbstractCollector {
                 source?: string;
             }[] = formatCaptures(tree, publicAccessorsCaptures);
 
-            console.log("public accessors captures", accessorsTextCaptures);
+            dlog("public accessors captures" + accessorsTextCaptures.toString());
```
BridgeAR, line 48: Debuglog uses the same argument handling as console.log. Was there a strong reason to change this?
That question applies to all of these changes.

BridgeAR (approved):
LGTM % nits.

src/parser/resolver/callExpressions/AbstractCollector.ts
ResistantBear: Mh, it did actually not work to simply replace console.log with debuglog in any case, as debuglog expects a string as first parameter, and console.log was sometimes used with just an object passed as parameter, which does not work with debuglog.

BridgeAR: The implementation that is triggered in Node.js passes the arguments through the same code paths. That should therefore not be different.

image

ResistantBear: Mh, ok. I don't know. I just got the error from the image below when trying to build the project with just replacing `console.log` with the debug logger. According to the IDE, the latter is defined as 
`export type DebugLoggerFunction = (msg: string, ...param: unknown[]) => void;`
while console.log is has the type
`log(message?:` any, ...optionalParams: any[]): void;`
which also accepts other types than "string" as first parameter.

grafik

BridgeAR: The types are no official documentation of the functionality and the type definition for debuglog is just wrong. Looking at the official Node.js documentation, it is actually not documented at all how things are printed. Opening a PR to Node.js to improve that would be nice. To do so, I suggest to check the actual implementation of the feature.

Type definitions are inevitable faulty. They are defined by humans and it will therefore for certain deviate from the actual implementation from time to time. A great example where the type definition is likely always going to be somewhat wrong is `JSON.stringify()`. It has some functionality that is just very hard to write a type definition for.

In case of type errors you may therefore not rely upon them being correct using TypeScript.

ResistantBear: Interessting. 
For now, I just restored the old parameter passing where it was not strictly required to satisfy the type system, which is the case mentioned above.

Issue #18, opened 2023-12-22
Lines of code not always correct


ResistantBear:
I have tested a C++ source code file with 739 lines of code. However, the json produced by metric gardener states that the file has 0 lines_of_code and 0 real_lines_of_code, which is just not correct.

ResistantBear:
Ok, the root cause for this issue seems to be that for some files, the root node of the tree returned by the parser of Tree-sitter is "ERROR" instead of "translation_unit", which is probably a bug of the C++ parser of Tree-sitter. 

However, I wonder if it is necessary to check if there is such a node? Are there cases where this procudure solves an issue? Otherwise, I would suggest to remove this procedure in favor of returning `tree.rootNode.endPosition.row` as the lines of code, which is robust in the case of an deviant root node. It should improve performance, as it does not need to perform any queries, too.

ce-bo:
Is there no translation unit expression in the grammar anymore?

I like your suggestion to just use tree.rootNode.endPosition.row instead. Do you know how the LOC metric is implemented in metric-gardener? Is it just the number of lines in a file or is it the number of lines minus e.g. empty or comment lines?

The Definition of LOC in metric gardener should be explained somewhere. Would you take care please?

ResistantBear:
I already asked myself if there is a concrete definition of the available metrics somewhere, as I know that one can define "lines of code" in many different ways... I assume there is no such definition documented currently? 
As far as I can say, the current code just returns the overall number of lines for LOC, while there is the separate "Real LOC" metric for ignoring comments and empty lines.

There is still a translation_unit expression and the current procedure seems to work fine for most files. However, as mentioned, for some files, Tree-sitter generates an "ERROR" node instead, which is probably, well, an error 😄

ce-bo:
I think there is no description of metrics yet. You might want to add it to the Readme?

PR #21, opened 2023-12-28
Fixed calculation of lines of code, added test cases

ResistantBear:
Replaced the old query for a start node with calculating the lines of code directly from the root node, which improves reliability and performance as discussed in [#18](https://github.com/MaibornWolff/metric-gardener/issues/18#issuecomment-1870244297).

Additionally, this fixes an off-by-one issue when calculating the lines of code. Until now, there was added one to the result only if there is an empty line at the end of the file. However, there has to be also added one if there is an non-empty last line of code. This change should also be benefitial for the performance. 

Compared to the main branch, these changes improve the performance of analyzing the JUnit 4 code base from around 10,5 s to 9,5 s.

Added a description of the lines of code and real lines of code metric to the Readme, based upon the source code.

Added test cases for the lines of code metric for the three languages for which there are existing test cases for other file metrics (PHP, TypeScript and Go). In order to have test files which have no empty line at the end, etc., I had to add the resources-folder to the .prettierignore-file in order to avoid that this formatter changes these aspects of the test files.

Introduced the changes to lines of code to the real lines of code calculation and added test cases for real lines of code (until now there are only test cases for TypeScript, as there is an unresolved issue with this metric outside of the scope of this request that should be clarified, see #22).


BridgeAR (approved):
LGTM % nit.

We can also improve the `countEmptyLines()` method https://github.com/MaibornWolff/metric-gardener/pull/21/files#diff-beb7e726fc334187372ca48c0fe0ded66e194c6b0dbc98e7b60aeb547a996ea9R69.

E.g.,
`text.match(/^\s*$/gm).length` should also work.

ResistantBear:
> LGTM % nit.
> 
> We can also improve the `countEmptyLines()` method https://github.com/MaibornWolff/metric-gardener/pull/21/files#diff-beb7e726fc334187372ca48c0fe0ded66e194c6b0dbc98e7b60aeb547a996ea9R69.
> 
> E.g., `text.match(/^\s*$/gm).length` should also work.

I tried out this suggestion as it seems a good idea.
However, it delivers strange results, counting almost any line of the file `./resources/typescript/real-lines-of-code.ts` as empty. Even more weird, it does work just fine when copying the files content as constant string into the source code. I have tried many things up to printing out the char codes to analyze the differences of both inputs, but the only difference between them lies in the string from parsing the file having Windows-style line endings (\r\n), while the constant string has Unix-Style endings (\n). However, using `/^\s*\r?$/gm` does not work, either.

PR #29, opened 2024-01-04
Replaced Regex for empty comment lines

ResistantBear:
Instead of counting every comment line as empty that does not contain any alphanumeric characters, this now looks for any characters that are not used to mark comment lines following the inline comment line syntaxes listed in [this paper](https://www.researchgate.net/publication/335541386_TAXONOMY_AND_DESIGN_CONSIDERATIONS_FOR_COMMENTS_IN_PROGRAMMING_LANGUAGES_A_QUALITY_PERSPECTIVE),
the C-style block comment syntax or the "%" syntax, used, e.g., by latex. 

However, if a non-empty comment line includes only symbols that are included in the comment syntax for one of the supported languages, these lines are still incorrectly counted as empty. I do not really see how to improve this, unless performing distinction of cases based on the programming language.

Further changes:

- Added code documentation and test cases for the comment_lines metric.

### DIFF
--- a/src/parser/GenericParser.test.ts
+++ b/src/parser/GenericParser.test.ts
@@ -372,6 +372,30 @@ describe("GenericParser", () => {
         });
     });
 
+    describe("parses GO commentLines metric", () => {
+        it(
+            "should count number of comment lines correctly, including line with curly brackets, inline comments" +
+                " and text lines inside block comment",
+            () => {
+                const inputPath = fs.realpathSync(goTestResourcesPath + "if-statements.go");
+                const parser = new GenericParser(getParserConfiguration(inputPath));
+                const results = parser.calculateMetrics();
+
+                expect(results.fileMetrics.get(inputPath)?.get("comment_lines")?.metricValue).toBe(
+                    4
+                );
+            }
+        );
+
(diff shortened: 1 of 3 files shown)


ce-bo:
As discussed, let's simplify the calculation of the comment lines metric since we cannot handle all edge cases in a nice way.
The old logic has been added to mimic the metric definition of Sonar. Simply counting all comment lines (including empty comments) is fine for now. Further feature requests can be opened for implementing a more advanced metric (let's say "real_comment_lines").

ResistantBear:
Closed in favour of #33

ce-bo:
> > As discussed, let's simplify the calculation of the comment lines metric since we cannot handle all edge cases in a nice way. The old logic has been added to mimic the metric definition of Sonar. Simply counting all comment lines (including empty comments) is fine for now. Further feature requests can be opened for implementing a more advanced metric (let's say "real_comment_lines").
> 
> We haven't discussed whether to still exclude file header comments, as that works fine as far as I know. However, I guess that the new "simple" comment_lines metric should also count such comments?

I agree to just count them.

ce-bo:
@ResistantBear please remove the branch, if we can get rid of it.

Issue #24, opened 2024-01-02
Updating tree sitter grammars


ResistantBear:
Requested by @BridgeAR in https://github.com/MaibornWolff/metric-gardener/pull/17#pullrequestreview-1798071997

As there seem to be some issues with the existing mappings, too, these should be sorted out, too. There is, for example, only PHP as applicable language for all binary expression operators, or nodes marked as statements which are no statements, see https://github.com/MaibornWolff/metric-gardener/issues/30#issuecomment-1880841893

ce-bo:
Let's get rid of the current import script and introduce a new one that allows to update an already supported language and its already mapped expression easily.

ResistantBear:
Notes for further work:
- [ ] Binary_Expression gets removed in python, seems to use the syntax node boolean_operator instead. So special handling for binary_expression needs to be adjusted to also include this node. May change the code for the special handling to handle such cases more elegantly.
- [ ] Coupling test cases seem to be affected by updating the node mappings. That needs to be investigated.

Issue #26, opened 2024-01-02
Snapshot test cases fail because the generated JSON contains the absolute file paths


ResistantBear:
Snapshot test cases fail because the generated JSON contains the absolute file paths to the parsed files on several places, which has of course a different prefix compared to `"C:\Users\CedrikB\projects\node-parser"` in case someone else executes them. I guess this is unavoidable as long as the absolute file path is included in the JSON and the test case compares the complete JSON output. 

_Originally posted by @ResistantBear in https://github.com/MaibornWolff/metric-gardener/pull/17#discussion_r1437609545_

PR #30, opened 2024-01-08
Remove bruteforce method and decoupled file extension from file language

ResistantBear:
- Removed `QueryBuilder.getBruteForcedStatementsQuery()` which is responsible for most of the execution time of this project. This reduces the time to analyze the JUnit 4 Codebase from ~ 8,8 s to ~ 2,3 s. It works fine with all tested languages, delivering identical results in the test cases and for analyzing the JUnit 4 codebase, respectively.

- Decoupled file extension from the language of the file, as there can be multiple file extensions for files of a single language (e.g. .cpp, .cc, .c, .hpp, .h, etc. for C++)

The failing of all test cases for PHP was due to `TSQueryErrorNodeType` errors being thrown. This was caused by some errors in the expression mappings regarding the node "function" which were unnoticed until now, since the old bruteforce approach tested compatibility with the selected language by trying to run a query with every single available statement and catching the error that is thrown if the statement is incompatible with the language. Additionally to statements being included that are not supported, there are also supported statements being excluded by the expression mappings that were used before with the bruteforce method.

I compared the chosen statements by the old and the new approach and manually adjusted the node mappings for "function" by marking this node as not applicable to PHP, which makes the test cases run again. 
The new import script for #24 should solve this issue.


ResistantBear:
> If I understand correct, the current implementation is responsible for identifying if a method can be used for the code or not? I would definitely not want to have predefined feature sets per language. Instead, we should have a feature detection which only runs a query once to see if it may apply or not. It should not be relevant if a feature is handled identical in programming language a and b of if it is not. That way, we keep the flexibility (which we should increase instead of decreasing it) and we gain performance by not running queries that are not supported more than once per language (we might need an intermediate state. It could be that a query works on some code but not on all of a programming language).

Well, its difficult... the current implementation does check whether the statement is applicable to a language without relying on an entry for this in the node mapping file. But this entry in the mapping file is set automatically by parsing the node_types.json which comes with each tree-sitter grammar. So the only difference should be that we now exclude statements for which we already know that they are not supported by the corresponding grammar of the language, instead of testing each statement type that any language has and checking if that throws an exception, which gets thrown by tree-sitter if the grammar of the language does not support the statement. Which again requires the language grammar, of course. We cannot parse anything without a grammar. 

At least in theory. In practice, I have found some issues with this, however, they are not really related to this new approach of filtering out statements, but seem an issue related to the parsing of these node_types.json done by src/commands/treeSitterMapping.ts. This method does classify too much nodes as a "statement", including things like "null" or "integer_literal" or the "function" keyword in the case of PHP, which is not correct und caused this issue.

src/parser/queries/QueryBuilder.ts
```diff
@@ -26,50 +34,114 @@ export class QueryBuilder {
...
+        return statementQueries.join("\n");
     }
 
-    // This is not necessary anymore
-    // due to the mapped node types (/resources/node-types-mapped.config)
-    // and the information per expressions
-    // in which language it is included
-    // Thus, we do not have to figure out
-    // if the expression is included in a language by query the expression
-    // and catch an exception if it is not included.
```
ResistantBear, line 49: In general, this change implements the removal of the bruteforce approach already planned by @ce-bo as stated in this comment.

BridgeAR:
My point is: we should support programming languages that use the same tokens as our statements without officially supporting it.

Let's say programming language A uses tokens a1, a2, b1, c1, while language B uses tokens a1, b1, c2.
The user wants to parse language C. Our metrics now need tokens a1 and b1 for the concrete metric and are able to understand how to handle the tokens even though we do not have official support for language C.

ResistantBear:
> My point is: we should support programming languages that use the same tokens as our statements without officially supporting it.
> 
> Let's say programming language A uses tokens a1, a2, b1, c1, while language B uses tokens a1, b1, c2. The user wants to parse language C. Our metrics now need tokens a1 and b1 for the concrete metric and are able to understand how to handle the tokens even though we do not have official support for language C.

I understand that. However, in order for that language C to be handled, there needs to be at least a parser/grammar for that language to be present, otherwise, we are unable to obtain a syntax tree for a text in that language in the first place. In that case we can still precompute a file stating which statements are available for any language for which there are parsers available.

Maybe we could also include the old "brute force and exception catching" method to this automatic precomputation, if it is not possible to automatically parse that out of the node_list itself without the current errors regarding the classification as a  "statement"  (though I hope it is possible). 
We could also follow your suggestion and execute the bruteforce check every startup for every language for which a parser is installed or something like that. Or have a procedure where such precomputations can be executed by the end user after installing an additional parser while keeping the manual mappings for the nodes relevant for our metrics. I guess such approaches should work as long as there are no new/different node types included in the grammar of that language that are relevant for our metrics.

BridgeAR:
That is indeed something we should look into. Users should not be required to load parsers but we could include the known parsers, even if we do not yet support the language to theoretically support it. That would be a good option for me as long as the individual language parsers are not too huge (which I hope they are not).

src/parser/helper/Model.ts
```diff
@@ -51,23 +88,43 @@ export class ExpressionQueryStatement implements QueryStatementInterface {
...
+            }
+        }
+
+        this.applicableForLanguages = new Set();
+        for (const languageString of applicableForLanguages) {
+            const language = languageToAbbreviation.getKey(languageString);
+            if (language !== undefined) {
+                this.applicableForLanguages.add(language);
+            }
+        }
```
ce-bo, line 123: Let's refactor this duplicated code.

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -23,20 +23,21 @@ export class RealLinesOfCode implements Metric {
      * @param allNodeTypes List of all configured syntax node types.
      */
     constructor(allNodeTypes: ExpressionMetricMapping[]) {
+        // Should I use this? What about the category == "comment" thing?
+        // getQueryStatements(allNodeTypes, this.getName());
```
ce-bo, line 27: Please fix this unresolved todo

src/parser/queries/QueryBuilder.ts
```diff
@@ -26,50 +34,114 @@ export class QueryBuilder {
...
-                continue;
+    /**
+     * ONLY FOR DEBUGGING PURPOSES!
+     * This lets you compare the result of the new approach for getStatementsQuery which relies on the "language"
+     * field of the expression mappings to the old, very slow bruteforce approach.
+     * @param returnBruteforceResult Whether to return the result of the bruteforce approach or the new approach.
+     * @return A query string.
+     * @private
+     */
+    private debugCompareBruteforceToMapped(returnBruteforceResult: boolean) {
```
ce-bo, line 84: I know it is still a draft one, but don't forget to resolve this.

src/parser/helper/Model.ts
ResistantBear: Should be done now.

src/parser/metrics/RealLinesOfCode.ts
ResistantBear: I have removed that, as it is resolved in #32

src/parser/queries/QueryBuilder.ts
```diff
@@ -26,50 +34,114 @@ export class QueryBuilder {
...
+     * Returns an empty query if the language is not supported.
+     * @return A query string.
+     * @private
+     */
+    private getStatementsQuery() {
+        const statementQueries: string[] = [];
 
-        return new Query(this.treeSitterLanguage, queryString);
+        for (const statement of this.statements) {
+            // Is "activatedFor" still necessary? I have only found one node with the activated_for_languages filed.
```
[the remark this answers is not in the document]
ResistantBear: All right, I have removed the comment.

src/parser/queries/QueryBuilder.ts
ResistantBear: Yes, we might remove this once the issue with the node mappings is resolved. Until then, it is quite helpful to compare the results of the old and the new approach if necessary.

I don't remember if we discussed how to proceed on this issue. Should we first implement a new procedure for updating the grammars and generate a new node mappings file which fixes the issues that occured here? Or should I try to fix the existing file until the output from the dependency analysis equals that of the test case (minus the thing mentioned in #26) ?

ResistantBear: Debug bruteforce function removed

ResistantBear:
> That is indeed something we should look into. Users should not be required to load parsers but we could include the known parsers, even if we do not yet support the language to theoretically support it. That would be a good option for me as long as the individual language parsers are not too huge (which I hope they are not).

That should be tried then once the new import script (#43) and this is merged. I have opened #52 to take a look on that.

mylinhdao (approved):
LGTM after changing function and variable names.

Issue #31, opened 2024-01-09
Comment blocks in python are not handled


ResistantBear:
Comment blocks in python like
"""
comment line here
"""
are not handled. 

This is because such comments get parsed as (expression_statement (string)) instead of (comment). As that seems to be an issue with the tree-sitter parser for python, I do not recommend to fix this here, as that would necessitate a workaround especially for that language.

BridgeAR:
This is a tricky one. Python handles multi-line comments different than other programming languages. They are just "regular" expressions.
How to interpret them correct is a difficult decision and I guess that won't be "fixed". As such, having a special handling in our code might be required to parse those. Alternative print a warning when parsing python with the concrete metric.

BridgeAR:
You might want to open an issue in the python library to ask if they would be interested in trying to differentiate the two or not.

ResistantBear:
> This is a tricky one. Python handles multi-line comments different than other programming languages. They are just "regular" expressions. How to interpret them correct is a difficult decision and I guess that won't be "fixed". As such, having a special handling in our code might be required to parse those. Alternative print a warning when parsing python with the concrete metric.

Ok, mh. Interesting. If they are indeed handled as normal expressions, the behaviour of classifying them as "expression_statement" should be correct then from a syntactic view, I guess. That would also arise the question, wheter or not to count them as "comment line" at all. If there is no strong opinion to count them into the comment lines metric, we might leave it as it is.

ResistantBear:
Such "afterMetricHooks" would also have to be specific for each metric. But otherwise, something similar should work, I guess. It might also happen that we need to override the general logic for some language one day because it is not sufficient to just add some other value to the result, or because the logic does not apply at all. On the other hand, if the general logic is ok and we only need to add to that result, like it is the case here, this solution could avoid redundancy. On the other hand, we could call the function with the general logic again from the overriding function to avoid that redundancy. Mh.

ResistantBear:
However, the main issue here would be that part of identifying these multi-line strings and determining that they are not passed to anything. I currently have a look at the dependency analysis code to fix an issue that occurs at identifying passed parameters, etc., when updating the grammars. There is a lot that can go wrong there when the grammar changes, also to keep up with all variants that the grammar allows for on such a position, so this is not going to be trivial and the maintenance of handling this could become an issue. But there is probably no way around it.

ResistantBear:
[...]

PR #32, opened 2024-01-09
New approach for calculating real lines of code

### LINKED ISSUE #22: Real lines of code not always correct
The real lines of code metric does not behave as expected when there is a comment in the same line as "real" code, e.g.:
`let a = 5; // Assigned 5 to a`
This can be also tested with one of the new test cases introduced by #21. 

I guess that this line counts as "comment line" and is therefore substracted from the number of lines of code, which is an approach that does not work in this case.

ResistantBear:
As discussed with @BridgeAR, I implemented a new approach for calculating the real lines of code metric.
This is based upon traversing the syntax tree instead of querying for comment lines and detecting empty lines and subtracting that from the total number of lines.
It does so by adding the first line of every non-comment syntax node to a set and calculating the length of that set.

Further notes: 
- Added test cases for this metric for the currently tested languages PHP, TypeScript and Go. 
- Added first test cases for python, as that languages uses a quite different syntax for code blocks than the C-style syntax used by the tested languages. This change works fine, but there occured a problem with block comments. As that is not related to this new computation approach for real lines of code, I have opened issue #31 for this.

Resolves #22


src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -16,75 +12,75 @@ let dlog: DebugLoggerFunction = debuglog("metric-gardener", (logger) => {
...
-            grammars.get(parseFile.language),
-            tree,
-            parseFile.language
-        );
-
-        queryBuilder.setStatements(this.commentStatementsSuperSet);
+        const cursor = tree.walk();
+        // Assume the root node is always some kind of program/file/compilation_unit stuff
+        cursor.gotoFirstChild();
+        this.walkTree(cursor, sureCodeLines);
```
BridgeAR, line 70: What about letting the first method call to set the set variable and to let the method return that?
That way we do not change an outside variable inside the method.

```suggestion
       const actualCodeLines = this.walkTree(cursor);
```

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -16,75 +12,75 @@ let dlog: DebugLoggerFunction = debuglog("metric-gardener", (logger) => {
...
+     * Uses a {@link TreeCursor} for this, as according to the
+     * {@link https://tree-sitter.github.io/tree-sitter/using-parsers#walking-trees-with-tree-cursors|Tree-sitter documentation},
+     * this is the most efficient way to traverse a syntax tree.
+     * @param cursor A {@link TreeCursor} for the syntax tree.
+     * @param sureCodeLines A set in which the line numbers of the found code lines are stored.
+     */
+    walkTree(cursor: TreeCursor, sureCodeLines: Set<number>) {
+        // This is not a comment syntax node, so assume it includes "real code".
+        if (!this.commentStatementsSet.has(cursor.currentNode.type)) {
+            // Languages like Go do use new line characters for marking the end of a statement, which leads to wrong
```
BridgeAR, line 39: Would this be something that we could fix in the tree sitter go package? As in: is that something they might also be interested in changing?

ResistantBear: No, I think that works as intended. While \n has no meaning in other languages and is therefore left out of the syntax tree with the other languages, in Go it seems to be more or less equivalent to a semicolon, so it should be part of the syntax tree.

But as I think of it, since I have removed the inclusion of the last row into the set, I could also remove handling this special case, as it only affected the value of the last row (being one more, due to \n spanning also into the next line).

src/parser/GenericParser.test.ts
```diff
@@ -266,12 +289,12 @@ describe("GenericParser", () => {
 
     describe("parses TypeScript real lines of code metric", () => {
         it("should count correctly for a non-empty file, ignoring comments and empty lines", () => {
-            const inputPath = fs.realpathSync(tsTestResourcesPath + "real-lines-of-code.ts");
+            const inputPath = fs.realpathSync(phpTestResourcesPath + "php-example-code.php");
```
ce-bo, line 292: I am confused by this change since the description is related to TypeScript :-)

src/parser/metrics/RealLinesOfCode.ts
ResistantBear: So I should introduce something like the following?
`function walkTree1(cursor): Set<number>{
    let sureCodeLines = new Set();
    this.walkTree(cursor, sureCodeLines);
    return sureCodeLines;
}`

Otherwise I would have to instantiate a new set with every recursive call and merge that after returning from each recursive call, which I guess will be quite inefficient compared to now?

src/parser/GenericParser.test.ts
ResistantBear: Ah, yes, that is wrong, thank you. It should be corrected now.

This long file starts to get confusing, especially when merge conflicts occur. I will split it into separate files one the "merging nightmare" #23 is merged (introducing "async" and "await" in every test case).

src/parser/metrics/RealLinesOfCode.ts
BridgeAR: I meant something like this:

```js
function walkTree(cursor, sureCodeLines = new Set()) {
  // regular walkTree logic
  return sureCodeLines;
}
```

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -16,75 +12,70 @@ let dlog: DebugLoggerFunction = debuglog("metric-gardener", (logger) => {
...
+     *
+     * Uses a {@link TreeCursor} for this, as according to the
+     * {@link https://tree-sitter.github.io/tree-sitter/using-parsers#walking-trees-with-tree-cursors|Tree-sitter documentation},
+     * this is the most efficient way to traverse a syntax tree.
+     * @param cursor A {@link TreeCursor} for the syntax tree.
+     * @param sureCodeLines A set in which the line numbers of the found code lines are stored.
+     */
+    walkTree(cursor: TreeCursor, sureCodeLines: Set<number>) {
+        // This is not a comment syntax node, so assume it includes "real code".
+        if (!this.commentStatementsSet.has(cursor.currentNode.type)) {
```
BridgeAR, line 38: There might also be "whitespace" nodes. If those exists, we should also exclude those.

src/parser/metrics/RealLinesOfCode.ts
ResistantBear: Ah, yes, that looks more elegant. I have changed the code accordingly.

src/parser/metrics/RealLinesOfCode.ts
ResistantBear: I have traversed through some trees by now, and never encountered such a node. Not even with python, where I expected them most because of them having a semantic meaning. Seems like the tree-sitter grammars abstract them away when parsing the syntax tree.

ResistantBear: It seems like handling whitespace-sensitive languages with tree-sitter is a more complicated topic, as all grammars that parse them have to rely on external software to handle them (https://github.com/tree-sitter/tree-sitter/issues/219). Which might explain why they are not included in the resulting tree.

BridgeAR: Thank you for looking that up! That is good to know.

BridgeAR (commented):
This looks really good! It simplifies the overall logic a lot.

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -16,75 +12,70 @@ let dlog: DebugLoggerFunction = debuglog("metric-gardener", (logger) => {
...
+     * @param cursor A {@link TreeCursor} for the syntax tree.
+     * @param sureCodeLines A set in which the line numbers of the found code lines are stored.
+     */
+    walkTree(cursor: TreeCursor, sureCodeLines = new Set<number>()) {
+        // This is not a comment syntax node, so assume it includes "real code".
+        if (!this.commentStatementsSet.has(cursor.currentNode.type)) {
+            // Assume that first and last line of whatever kind of node this is, is a real code line.
+            // This assumption should hold for all kinds of block/composed statements in (hopefully) all languages.
+            sureCodeLines.add(cursor.startPosition.row);
+            // sureCodeLines.add(cursor.endPosition.row);
```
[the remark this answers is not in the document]
ResistantBear: It does work as expected. I have tried quite hard to break it in the unit tests with the file resources/typescript/weird-lines.ts, but it does indeed also work with code like this:

````ts
for (i =
          0

     ;
      i < 100;
      i++) {
     // do something
     let a =
         5;
     a++ // a gets incremented
     ;
}

do {
     // do something
} while (x < /* x has to stay lower than 100 */ 100);
````

Adding all lines between start and end of each syntax nodes would
include thinks like empty lines inside code blocks. Or, in case of the
root node, even all lines of the whole file, so that does not work.

I first thought that I need to include at least the last row of a syntax node (end of a code block, etc.). However, as this approach traverses the concrete syntax tree, also including  any kind of "unnamed" syntax nodes like "{" or "(", any node that represents syntax that could mark the end of such a block is also visited, whether it being some kind of bracket or a last statement of a block (like in python).
This code would fail if there is some kind of syntax node that a) spans over multiple non-empty lines and b) does so without having child nodes in the syntax tree that start on these other, non-empty lines. I actually do not think that such a node exists (other than the "comment"-nodes, which should not be counted), as every non whitespace/non-control-character token of code that is relevant for the semantic of the program ("real code") seems to be included in the syntax tree, and I have no idea of a token of code that could span over multiple lines, as long as it is not allowed to have, for example, variable names like `"hello\nworld"` in a language.

TLDR: This surprisingly simple implementation "just works" :wink:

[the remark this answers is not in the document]
ResistantBear: Thinking of it, I could check for each node if it is a leaf node and if it does span over more than one line to handle such a case. Not sure if that would be worth the extra computing expense.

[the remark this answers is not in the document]
BridgeAR: It is a great implementation! I can't think of anything going wrong either. Let's work with this for now and if we find cases that need a different handling, we can still change it.

BridgeAR (approved):
LGTM. It just needs a rebase/merge main :)

src/parser/metrics/RealLinesOfCode.ts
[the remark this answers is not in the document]
ResistantBear: Good. I will merge it once this PR is approved.

PR #33, opened 2024-01-10
Redefine comment_lines metric

### LINKED ISSUE #28: Comment Lines is not always correct
When testing the file `./resources/php/php-example-code.php` for comment lines, the result is 7, but should be 8.

ResistantBear:
In accordance to https://github.com/MaibornWolff/metric-gardener/pull/29#issuecomment-1883288691 and assuming that the answer to https://github.com/MaibornWolff/metric-gardener/pull/29#issuecomment-1884883032 is "yes", this redefines the comment_lines metric to include all comment lines, regardless of whether they are empty, at the beginning of a file, etc..

Additional notes:
- Removes now-obsolete code
- Adjusts result and description of the test cases for the comment_lines metric where necessary
- Adjusts documentation in JSDoc and Readme accordingly.

Resolves #28


src/parser/metrics/CommentLines.ts
```diff
@@ -37,21 +39,12 @@ export class CommentLines implements Metric {
...
-            const commentLinesWithText = this.findTextLines(captureNode.text).length;
-            const docBlockLines = this.findDocBlockLines(captureNode.text).length;
-
-            return accumulator + commentLinesWithText - docBlockLines;
-        }, 0);
+        let commentLines = 0;
+        for (const match of matches) {
+            const captureNode: SyntaxNode = match.captures[0].node;
+            commentLines =
+                commentLines + 1 + (captureNode.endPosition.row - captureNode.startPosition.row);
```
BridgeAR, line 46: What about:

```
if (answer === 42 /* abc */ && wrong === 'never' /* second comment */) {
```

ResistantBear: Ah, yes, that will fail because there are multiple nodes for the same line. How embarassing. 
I guess I will change that to a set of lines, similar to #32.

BridgeAR: You only have to memorize the last line. If the start of the next one is the same as the end of the last, substract one.

ResistantBear: Mh, that could also work, yes. 
However, I am unsure if I can rely on that the results/matches of the query are ordered regarding the lines of code.

BridgeAR: I am confident that is the case, you can also verify that with a test and use the faster implementation as long as the test is green.

README.md
```diff
@@ -57,8 +57,8 @@ This is quite slow and can take up to one or two hours but can provide good resu
 -   mcc
 -   functions
 -   classes
--   lines_of_code - The number of lines inside a file
--   comment_lines (not unit tested)
+-   lines_of_code - The number of lines of a file
+-   comment_lines - The number of comment lines inside a file
```
ce-bo, line 61: Let's define the metrics more precisely.
It should be clear as much as possible what a metric is counting:
- Lines of Code = The number of lines of a file including empty lines.
- And comment_lines = the number of comment lines inside a file. Doc Blocks, empty comment lines, comments in same line as code, etc.

I think we should optimize this for all metrics (maybe an a follow up PR) by providing proper metric definitions.

ce-bo (approved):
LGTM but PTAL at one minor remark.

src/parser/metrics/CommentLines.ts
ResistantBear: Seems to work with the less computational-intensive solution suggested by you, so I have changed that accordingly

PR #34, opened 2024-01-12
Use project location and platform-idependent file paths in test cases, further fixes for snapshot tests.

### LINKED ISSUE #25: Remove seemingly obsolete field "types" in the json file metrics output
> Regarding the types field, as this is always empty, I don't know what it is used for. I will look into the source code to find some hints about it.

I have found in the Git history that @ce-bo has added this "types" field in src/commands/outputMetrics.ts to store the "namespace" coupling metric in the commit https://github.com/MaibornWolff/metric-gardener/commit/e134ae4c7be9455262dee4b4719671c1a1c08134
However, this whole namespace metric and the code to add it to the field has been removed in https://github.com/MaibornWolff/metric-gardener/commit/2becab6f8d454c8984d84f45c262eaeee8f9a1d5, while the field "types" still remains.
I guess that we should remove the now unused field "types" again?

_Originally posted by @ResistantBear in https://github.com/MaibornWolff/metric-gardener/pull/17#discussion_r1438104275_

### LINKED ISSUE #26: Snapshot test cases fail because the generated JSON contains the absolute file paths

ResistantBear:
Makes the dependency analysis snapshot test cases project location and platform-independent by using paths relative to the given sources directory and replacing every slash by a backslash as Windows does not allow forward slashes in paths, while unix migh allow backslashes. Also adds sorting of coupling results before the snapshot comparison, as the order in which files are found (and added to the results, accordingly) seems to be different on MacOS compared to Windows/Linux.
Also removes the obsolete field "types" from the json output.

Updates all snapshot files be compatible with these changes and modifies gitignore to include snapshot files for the compiled .js code.

Confirmed to run under MacOS and Linux, tests outstanding for Windows.

This is based upon the new file handling itroduced by #23 and should not be merged before that pull request is approved, so this has to stay a draft until #23 is merged.

Resolves #26
Resolves #25


src/parser/metrics/coupling/Coupling.ts
```diff
@@ -95,6 +99,18 @@ export class Coupling implements CouplingMetric {
...
+            for (const relationship of relationships) {
+                relationship.fromSource = formatPrintPath(relationship.fromSource, this.config);
+                relationship.toSource = formatPrintPath(relationship.toSource, this.config);
+            }
+            const relCouplingMetrics = new Map();
+            for (const [absolutePath, metricValues] of couplingMetrics) {
+                relCouplingMetrics.set(formatPrintPath(absolutePath, this.config), metricValues);
+            }
+            couplingMetrics = relCouplingMetrics;
+        }
```
ce-bo, line 112: Let's extract this into a separate method.

src/parser/helper/Helper.ts
```diff
@@ -12,50 +14,122 @@ export function formatCaptures(tree, captures) {
...
+ * @return negative value if a < b, a positive value if b < a, and 0 if a === b.
+ */
+export function strcmp(a: string, b: string) {
+    if (a < b) {
+        return -1;
+    } else if (b < a) {
+        return 1;
+    } else {
+        return 0;
     }
```
ce-bo, line 31:
```suggestion
    if (a < b) {
        return -1;
    }
    if (b < a) {
        return 1;
    }
    return 0;
```

ce-bo (approved):
LGTM

src/parser/metrics/coupling/Coupling.ts
ResistantBear: Done

src/parser/helper/Helper.ts
ResistantBear: Done

Issue #41, opened 2024-01-23
Add definition of "McCabe


ResistantBear:

BridgeAR:
The issue is that each programming language has some special handling and it is not as formalized due to that. Mostly, each implementation slightly deviates from each other. We can orient ourselves on the original suggestion as outlined in e.g., Wikipedia and see how it deviates to e.g., Sonar.

ResistantBear:
Further discussion about MCC:
https://github.com/MaibornWolff/metric-gardener/pull/42#discussion_r1469558033

ResistantBear:
Reading the definition used in https://docs.sonarsource.com/sonarqube/latest/user-guide/metric-definitions/ and having in mind the original defintion and the intention of @ce-bo to still include things like logical operators into the metric, which SonarQube does but the original definition of MCC, which is solely based on the control tree of a program, does not, I think that calculating the cyclomatic complexity per function in a similar way to SonarQube and using the maximum of the measured complexities inside one file as a "maximum of the cyclomatic complexities in the file" should be a viable option. As far as I understand, CodeCharta does also show such a value as "MCC" currently, with MCC meaning "maximum cyclomatric complexity" instead of McCable Complexity. By traversing the tree instead of using queries and resetting the count of "things that increase the complexity" when encountering a new function definition, that should be straight forward to implement.

On the other hand, we could also continue to just count these "things that increase the complexity" per file, which might give a better quantitative overview about the amount of logic inside the file, but less information about the complexity of the functions within it, as 100 functions with one if-statement each would have a really high value in this metric, but might not be that hard to understand. We could also include both options as separate metrics.

ce-bo:
The current approach for calculating the MCC metric is similar to the one of Sonar. It is based on the number of paths through the source code but also counts (as you already mentioned) operators like &&, ||, etc. and control structures like if, for, while, etc.
Also foreach is a control structure and should be counted as well.

I would like to rename the metric to "complexity" and add a definition to the Readme like I just did above.

ResistantBear:
> The current approach for calculating the MCC metric is similar to the one of Sonar. It is based on the number of paths through the source code but also counts (as you already mentioned) operators like &&, ||, etc. and control structures like if, for, while, etc. Also foreach is a control structure and should be counted as well.
> 
> I would like to rename the metric to "complexity" and add a definition to the Readme like I just did above.

[...]

PR #42, opened 2024-01-23
Add unit tests for Java, split test cases per language

### LINKED ISSUE #40: Add tests for file metrics for Java
- Test the metrics returned by calculateMetrics(...) in GenericParser for every metrics (lines_of_code, real_lines_of_code, commentLines, classes, functions, mcc)

ResistantBear:
Added unit tests for basic file metrics for java source code. Also introduces a new function to reduce code duplication inside the test case. Splits the test cases of GenericParser.test.ts into multiple test files, to have one test file for each programming language.

MCC is not yet tested due to the unclear definition of that metric (see #41).

Main author @mylinhdao 

Closes #40


resources/java/IfStatement.java
```diff
@@ -0,0 +1,25 @@
...
+        } else {
+            System.out.println(value + " is less than or equal to 0");
+        }
+        if (value > 0) {
+            System.out.println(value + " is greater than 0");
+        } else if (value < 0) {
+            System.out.println(value + " is less than 0");
+        } else {
+            System.out.println(value + " is equal to 0");
+        }
```
ce-bo, line 23:
```suggestion
        if (value > 0) return;
        if (value > 0) {
            System.out.println(value + " is greater than 0");
        }
        if (value > 0 && value > 1 ||  value > 5) {
            System.out.println(value + " is greater than 0");
        } else {
            System.out.println(value + " is less than or equal to 0");
        }
        if (value > 0) {
            System.out.println(value + " is greater than 0");
        } else if (value < 0) {
            System.out.println(value + " is less than 0");
        } else {
            System.out.println(value + " is equal to 0");
        }
```

test/parser/CSharpMetric.test.ts
```diff
@@ -0,0 +1,13 @@
+import fs from "fs";
+import { GenericParser } from "../../src/parser/GenericParser";
+import { getParserConfiguration, sortCouplingResults, testCouplingMetrics } from "./TestHelper";
+
+describe("C# metric tests", () => {
+    const csharpTestResourcesPath = "./resources/c-sharp/";
+
+    describe("parsing C# dependencies", () => {
```
ce-bo, line 8: @ResistantBear says: Wrong branch :-)

test/parser/GOMetrics.test.ts
```diff
@@ -0,0 +1,95 @@
+import { testFileMetrics } from "./TestHelper";
```
ce-bo, line 1: NIT: GO vs Go

test/parser/JavaMetrics.test.ts
```diff
@@ -0,0 +1,142 @@
+import { testFileMetrics } from "./TestHelper";
+const javaTestResourcesPath = "./resources/java/";
+
+describe("test metrics calculation for Java language.", () => {
+    describe("parses classes metric", () => {
+        it("count classes with different access-modifier", async () => {
```
ce-bo, line 6: Let's check the namings.

test/parser/JavaMetrics.test.ts
```diff
@@ -0,0 +1,142 @@
...
+                javaTestResourcesPath + "SingleCharacterClassName.java",
+                "classes",
+                2
+            );
+        });
+
+        it("dont count the fields or methods of class", async () => {
+            await testFileMetrics(
+                javaTestResourcesPath + "ClassWithFieldsAndMethods.java",
+                "classes",
```
ce-bo, line 29: Let's extract a const to define metric names on a central place once.

test/parser/JavaMetrics.test.ts
```diff
@@ -0,0 +1,142 @@
+import { testFileMetrics } from "./TestHelper";
```
ce-bo, line 1: Tests for comments are missing

test/parser/TestHelper.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+export async function testCouplingMetrics(inputPath: string) {
+    const realInputPath = fs.realpathSync(inputPath);
+    const parser = new GenericParser(getParserConfiguration(realInputPath, true, true));
+
+    const results = await parser.calculateMetrics();
+    const couplingResult = results.couplingMetrics;
+    sortCouplingResults(couplingResult);
+
+    expect(couplingResult).toMatchSnapshot();
```
ce-bo, line 65: Move this expect into actual test.

test/parser/TestHelper.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * Tesing if the file metric (mmc, loc, rloc, comments, classes, functions) is calculated correctly.
+ * @param inputPath path to testing source file
+ * @param metric can be: mmc, loc, rloc, comments, classes or functions
+ * @param expected expected test result
+ * */
+export async function testFileMetrics(inputPath: string, metric: string, expected: unknown) {
```
ce-bo, line 74:
```suggestion
export async function testFileMetrics(inputPath: string, metric: string, expected: number) {
```

test/parser/CSharpMetric.test.ts
ResistantBear: Belongs to issue #44

resources/java/IfStatement.java
ResistantBear: As this is used to calculate MCC, one could arise the question on how to handle your new example here regarding this metric. As effectively, returning immediately when the value > 0 would also render the other three if-loops basically useless, as the conditions of all of them will always be wrong. So the paths through this function effectively shrink to 3 here.

That is just a side node, as I do not suggest making the metric consider such things. To do such a thing, we would have to include a dead code detection or similar compiler-stuff.

resources/java/RealLineOfCodeAndComments.java
```diff
@@ -0,0 +1,55 @@
...
+        this.myString = initialString;
+    }
+
+    /**
+     * Getter method to retrieve the value of the `myString` variable.
+     *
+     * @return The current value of the `myString` variable.
+     */
+    public String getMyString() {
+        return myString;
```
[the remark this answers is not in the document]
ResistantBear: There is already a test file "InlineComment.java" to test inline comments in a separate test case, so yes, it is already covered.

resources/java/SingleCharacterClassName.java
```diff
@@ -0,0 +1,5 @@
+class I {
```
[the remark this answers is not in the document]
ResistantBear: Removed

resources/java/WhileAndForLoop.java
```diff
@@ -0,0 +1,12 @@
...
+    public void loopMethod(int n) {
+        for (int i = 0; i < n; i++) {
+            System.out.println("Iteration " + (i + 1));
+        }
+        int counter = 0;
+        while (counter < n) {
+            System.out.println("Iteration " + (counter + 1));
+            counter++;
+        }
+    }
```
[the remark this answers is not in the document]
ResistantBear: According to https://docs.sonarsource.com/sonarqube/latest/user-guide/metric-definitions/, SonarQube does increase the complexity by one for "if, for, while, case, &&, ||, ?, ->". That means that in your example, it would actually increase the complexity by one, but not because of the "foreach"-call itself, but because of the lambda function with the arrow operator. 

I guess that whole discussion should be part of #41, however

test/parser/JavaMetrics.test.ts
ResistantBear: Done

test/parser/TestHelper.ts
ResistantBear: Done

test/parser/TestHelper.ts
ResistantBear: Done

ce-bo:
I guess this is still in progress, so I moved it into the right state :-)

test/parser/JavaMetrics.test.ts
ResistantBear: Done with new FileMetric string enum

test/parser/JavaMetrics.test.ts
ResistantBear: Added now

ResistantBear:
We have encountered an issue where some tests "just fail" because of tree.rootNode being undefined when running all tests (both test.ts files and compiled .test.js files) e.g. via "npm test". This seems to be triggered by the higher number of test files or test cases in this PR, but also happened seldomly before. 

As it turns out, this is probably caused by a race condition in the tree-sitter library.
See https://github.com/tree-sitter/node-tree-sitter/issues/181
There is a potential fix for this linked, but it never got merged:
https://github.com/tree-sitter/node-tree-sitter/pull/75

We probably cannot do anything here to handle this issue on our side...

resources/java/IfStatement.java
ce-bo: It is just an example to demonstrate another type of if condition. Please adapt it.
The metric should not take care of dead code btw.

resources/java/WhileAndForLoop.java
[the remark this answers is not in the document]
ce-bo: This question is not blocking the extension of more cases for control structures in this code file.

ResistantBear:
> We have encountered an issue where some tests "just fail" because of tree.rootNode being undefined when running all tests (both test.ts files and compiled .test.js files) e.g. via "npm test". This seems to be triggered by the higher number of test files or test cases in this PR, but also happened seldomly before.
> 
> As it turns out, this is probably caused by a race condition in the tree-sitter library. See [tree-sitter/node-tree-sitter#181](https://github.com/tree-sitter/node-tree-sitter/issues/181) There is a potential fix for this linked, but it never got merged: [tree-sitter/node-tree-sitter#75](https://github.com/tree-sitter/node-tree-sitter/pull/75)
> 
> We probably cannot do anything here to handle this issue on our side...

It turned out that this issue is triggered especially if the number of test files is higher than the number of worker processes used by jest.
As a (temporal) mitigation for this issue, I reduced the number of test files by excluding the compiled .js files in the /dist/test folder from running with jest, as well as setting --maxWorkers to 24.

resources/java/IfStatement.java
ResistantBear: Done

mylinhdao (approved):
All remarks are solved.

Issue #39, opened 2024-01-22
May improve reliability of async operations


ResistantBear:
Is the following issue resolved for the asynchronous code handling the await of multiple promises in GenericParser and MetricCalculator? (see answers in the linked original pull request #23) 
If yes, this issue should be closed.

"The issue is, that this is a frequent pattern for an incomplete promise chain. Therefore, Node.js decided to track these promises to provide users a hint that they did not handle the promise properly. Node.js just can't know what is going to happen in the future and rather fails early."

_Originally posted by @BridgeAR in https://github.com/MaibornWolff/metric-gardener/pull/23#discussion_r1453521567_

ce-bo:
Since this is marked as further inquiry, would you add the open questions?

ce-bo:
@BridgeAR could you please give an advice on how to improve the promise chain handling? Thank you.

ResistantBear:
We finally clearified the possible pitfalls when working with lists of promises in Node.js with @BridgeAR.
Our current proposal in https://github.com/MaibornWolff/metric-gardener/pull/217 should be fine accordingly.

PR #43, opened 2024-01-24
New import script, update tree sitter grammars to newest available version

ResistantBear:
The import of syntax nodes from the installed tree sitter grammars now keeps the currently included mappings between syntax nodes and metrics specified in the current nodeTypesConfig.json-file instead of creating a blank file, reducing the maintenance work. 

In order to handle any breaking changes of the language grammars regarding the metric calculation process, the new script prints warning messages if a syntax node was removed from a language that was used to calculate a metric before. An overview over all syntax nodes which were added, removed or only for some languages added or removed from the grammars, a changelog with this information is written as a CSV-file by the import script.

Updating all tree sitter grammars and importing the syntax nodes with the new script works just fine for the file metrics, however, the substantial changes in the tree sitter grammar for PHP require changes in the source code of the coupling metric, which are not yet completed.


package.json
```diff
@@ -9,7 +9,7 @@
     },
     "scripts": {
         "start": "tsc && node dist/app.js",
-        "test": "tsc && jest",
+        "test": "jest",
```
ce-bo, line 12: tsc not needed anymore?

package.old.json
```diff
@@ -0,0 +1,60 @@
+{
```
ce-bo, line 1: PTAL

.gitignore
```diff
@@ -1,6 +1,5 @@
 .idea
 node_modules
-<<<<<<< HEAD
```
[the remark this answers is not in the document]
ResistantBear: Well, its actually a removed merge conflict. The merge conflict is in our current main. I probably messed that up with one of the earlier pull requests, oops.

package.json
ResistantBear: It is needed, thats the reason I included it here in an earlier PR. So this removal was an error, thanks for spotting it.

.gitignore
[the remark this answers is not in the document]
ce-bo: Ok, then please add another PR to fix this and rebase this branch afterwards

src/commands/Changelog.ts
```diff
@@ -0,0 +1,318 @@
+import { ExpressionMetricMapping } from "../parser/helper/Model";
```
ce-bo, line 1: In typescript, I am used to name files with a first lowercase letter.

src/parser/resolver/typeUsages/AbstractCollector.ts
```diff
@@ -197,6 +197,7 @@ export abstract class AbstractCollector {
                 const usedNamespace =
                     namespaceName + this.getNamespaceDelimiter() + nextUseItem.text;
 
+                // Why not just use nextUseItem.text here instead of split? It is set in the line directly above.
```
ce-bo, line 200: Good catch, feel free to adjust it.

src/commands/Changelog.ts
```diff
@@ -0,0 +1,318 @@
...
+        this.changelog.get(expression.expression)?.removedFromLanguage(languageAbbr);
+    }
+
+    /**
+     * Writes the changelog.
+     * @param metricMappings The current mappings to inform the user about which metric(s)
+     * are currently calculated with the changed syntax node(s).
+     * @return Promise that fulfills once the changelog has been written successfully.
+     */
+    writeChangelog(metricMappings: Map<string, ExpressionMetricMapping>) {
```
ce-bo, line 115: Let's split this function a little bit to be more readable.

.gitignore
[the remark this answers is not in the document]
ResistantBear: There is now #51 for this

src/commands/Changelog.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    /**
+     * Languages to which this node type was added.
+     */
+    addedLanguages: Set<string>;
+
+    /**
+     * Languages that include this node type and had included it before.
+     */
+    remainingLanguages: Set<string>;
```
ce-bo, line 275: It is not necessary to store remainingLanguages as a property here. It can be derived from old, removed and addedLanguages properties. providing a getter like `remainingLanguages()` would not force you to inject the needed data it in the constructor.

src/commands/Changelog.ts
```diff
@@ -0,0 +1,318 @@
...
+                    "Removed from language(s):" +
+                    csvSeparator +
+                    "Used for calculating the metric(s):" +
+                    csvSeparator +
+                    "Was explicitly only activated for language(s):" +
+                    EOL
+            );
+
+            for (const entry of this.getAll().values()) {
+                if (entry.remainingLanguages.size + entry.addedLanguages.size === 0) {
```
ce-bo, line 158: This could be a member function of ChangelogEntry. Pseudo Code:
```suggestion
                // Changelog class:
                // ....
                isExpressionRemoved() { return this.remainingLanguages.size === 0 && this.addedLanguages.size === 0; }
```

src/commands/Changelog.ts
```diff
@@ -0,0 +1,318 @@
...
+                    const mapping = metricMappings.get(entry.expression);
+                    if (mapping !== undefined) {
```
ce-bo, line 160: Let's change the condition to mapping === undefined and throw the exception immediately. Thus, you can remove the else.

src/commands/Changelog.ts
```diff
@@ -0,0 +1,318 @@
...
+                    csvSeparator +
+                    "Was explicitly only activated for language(s):" +
+                    EOL
+            );
+
+            for (const entry of this.getAll().values()) {
+                if (
+                    entry.remainingLanguages.size > 0 ||
+                    (entry.addedLanguages.size > 0 && entry.removedLanguages.size > 0)
+                ) {
```
ce-bo, line 209: Let's create a method to give this condition a readable name like `hasExpressionChanged()`. Is it correct to have remainingLanguages > 0 as an OR condition?

src/commands/Changelog.ts
```diff
@@ -0,0 +1,318 @@
...
+            }
+            writeStream.end();
+        });
+    }
+}
+
+/**
+ * Tracks the changes that were applied to one expression.
+ */
+class ChangelogEntry {
```
ce-bo, line 246: Since this class is not exported, the name should be fine. Otherwise I would rename it to express that this is more like an ExpressionChangelogItem or so.

src/commands/Changelog.ts
ce-bo, line 1: This is not related to the marked line. But let's think about the location of this file and the name. Maybe it should be located somewhere else e.g. `src/commands/import-grammars/ExpressionChangelog.ts`

src/commands/treeSitterMapping.ts
```diff
@@ -1,7 +1,18 @@
-import { binaryOperatorTranslations, ExpressionMetricMapping } from "../parser/helper/Model";
```
ce-bo, line 1: Let's imrpove the name of this file (`treeSitterMapping.ts`). Suggestion: `UpdateTreeSitterGrammars` or `TreeSitterGrammarUpdater`.

src/commands/treeSitterMapping.ts
```diff
@@ -13,77 +24,321 @@ const nodeTypeFiles = new Map([
...
+                                mapKey +
+                                '" was added for language "' +
+                                languageAbbr +
+                                '".'
+                        );
+                    }
+                }
+            }
+            continue;
+        }
```
ce-bo, line 206: Perfect candiate to be extracted into a separate method.

src/commands/treeSitterMapping.ts
```diff
@@ -13,77 +24,321 @@ const nodeTypeFiles = new Map([
...
+            dlog(
+                'New node type "' +
+                    nodeType.type +
+                    '" was added for language "' +
+                    languageAbbr +
+                    '".'
+            );
+        }
+
+        if (nodeType.subtypes !== undefined) {
```
ce-bo, line 241: I don't remember what a subtype is. Is it just a children of a node?

src/commands/Changelog.ts
ResistantBear: Done

src/commands/Changelog.ts
ResistantBear: Renamed to NodeTypesChangelog(Entry), just like the name of the file it produces

src/commands/Changelog.ts
ResistantBear: Mh, it would be oldLanguages - removedLanguages, basically. However, hat would force me to calculate the remaining languages each time I call that getter, and as remainingLanguages is used more than once, that would increase the computing effort. (This [Set difference operation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference) would be elegant in this case, but it does not seem to be available for anything but Safari yet 😉)

Injecting the data to the constructor should not be an issue as it is only performed by the Changelog-class inside the method `removedNodeFromLanguage` and sets it to the same value as `oldLanguages` there.

However, I notice that I do never use `oldLanguages`, so I could remove that filed from the ChangelogEntry. 
Even if one decides to include that in the CSV-file in the future, this would only be a one-time merge between remaining and removed languages (or just re-introduce the field oldLanguages).

src/commands/Changelog.ts
ResistantBear: Splitted into three new functions

src/commands/Changelog.ts
ResistantBear: It should be correct. Either there are remaining languages, in which case its clear that this is a modified node type and no completely removed or new one, or there are both added and removed languages, for the special case that every old language was removed but there were new languages added.

However, I changed that now to a more intuitive condition by using the filed `isNewNode`.

ResistantBear:
I have noted that the order of the new nodes being imported is not deterministic, which is probably caused by the asynchronous `updateLanguage` function, as the order in which languages are processed depends upon, for example, for which language the `readFile`-operation for the node-types.json-file completes first. This of course makes it harder to compare whether some changes to this script do lead to an equal result.

src/commands/treeSitterMapping.ts
ResistantBear: Mh, as it does not actually update the grammars themselves I have gone with "importNodeTypes", to make more clear that it imports the `node-types.json`-files and updates the `nodeTypesConfig.json`

src/commands/Changelog.ts
ResistantBear: I have created a import-grammars subdirectory and renamed the Changelog to NodeTypesChangelog.

mylinhdao (approved):
LGFM. After our refactoring

src/commands/treeSitterMapping.ts
ResistantBear: Splitted updateLanguage into separate methods

src/commands/treeSitterMapping.ts
ce-bo: Alright. Though, I would like to agree on a consistent naming for all these things again later. There should be a documentation about what a node type or an expression is exactly. I will create an issue for that.

src/commands/import-grammars/ImportNodeTypes.ts
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    return presentNodeTypesForLanguage;
+}
+
+async function updateLanguage(languageAbbr: string, presentNodes: Set<string>) {
+    const fileLocation = languageAbbreviationToNodeTypeFiles.get(languageAbbr);
+    if (fileLocation === undefined) {
+        console.error("No file path found for language " + languageAbbr);
+        return false; // Cancel whole update.
```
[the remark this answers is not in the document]
ResistantBear: And then use .catch(...) on the promise to handle errors? Just waiting for the Promise.all(...) for promises which might get rejected is no good idea based on https://github.com/MaibornWolff/metric-gardener/pull/23#discussion_r1452625942

See also #39

src/commands/import-grammars/ImportNodeTypes.ts
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    const nodeTypesPromise = fs.promises.readFile(fileLocation, "utf8").catch((reason) => {
```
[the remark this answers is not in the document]
ResistantBear: This script has to be executed when the tree-sitter grammars that metric-gardener uses are changed. 
So it is neither used when building nor during regular runtime, but manually when calling the program with the option --import-grammars.

src/commands/import-grammars/ImportNodeTypes.ts
```diff
@@ -0,0 +1,301 @@
...
+
+    const updatePromises: Promise<unknown>[] = [];
+
+    for (const languageAbbr of languageAbbreviationToNodeTypeFiles.keys()) {
+        let presentNodeTypes = languageToPresentNodeTypes.get(languageAbbr);
+
+        if (presentNodeTypes === undefined) {
+            presentNodeTypes = new Set();
+        }
+        const promise = updateLanguage(languageAbbr, presentNodeTypes).then(
```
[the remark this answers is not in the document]
ResistantBear: Restructured that part of the code in the meantime, there is now no .then anymore.
Removed the other .then for writing the file.

PR #48, opened 2024-01-25
Include unknown files

### LINKED ISSUE #36: Include unsupported files in the output
But we do we ignore files without file extension? Are they still highlighted in the overall files, just without any metrics? That would be my expectation. I would probably just return an empty string as extension and let the receiver handle that.
_Originally posted by @BridgeAR in https://github.com/MaibornWolff/metric-gardener/pull/23#discussion_r1452638134_
            
Note: Currently, all files with unknown or not present file extension are ignored, which has to be changed for this new feature. Handling file extensions is overhauled in https://github.com/MaibornWolff/metric-gardener/pull/30 to support multiple file extensions, so this change would have to be implemented on top of that branch or once that branch is merged.

ResistantBear:
Adds an array "info" into the json-output and lists all files with unknown file extensions there.

Additional improvements:
- Dependency analysis should now work with languages with multiple file extensions due to refactored file-language-detection

This builds upon the new file extension handling introduced by #30, so it should be only merged once this is approved.

Closes #36


ce-bo:
This seems to be blocked. Would you please mention why? Thank you.

ce-bo:
> Dependency analysis should now work with languages with multiple file extensions due to refactored file-language-detection

I think, the algorithm is not constructed to support this properly (yet). It might makes sense for detecting references between documentation and code files. More important is the dependency analysis within a set of code files that have the same programming language.

ResistantBear:
> This seems to be blocked. Would you please mention why? Thank you.

It was blocked because it was based on the new file handling introduced by #30, as mentioned in the top post of this PR.
As this is now merged, it is blocked because I should not review it myself

ResistantBear:
> > Dependency analysis should now work with languages with multiple file extensions due to refactored file-language-detection
> 
> I think, the algorithm is not constructed to support this properly (yet). It might makes sense for detecting references between documentation and code files. More important is the dependency analysis within a set of code files that have the same programming language.

I meant that it should now work if a programming language uses multiple file extensions for its source code, as it now no longer handles files based on file extension, but based on the language of the files. So we would no longer treat .cc and .cpp files as files of different languages. There is still no support to find dependencies between files of different languages, yes.

src/commands/outputMetrics.ts
```diff
@@ -11,6 +11,12 @@ interface OutputNode {
     };
 }
 
+interface OutputInfoNode {
```
mylinhdao, line 14: i recommend to rename OutputInfoNode to OutputUnknownFile.

src/commands/outputMetrics.ts
```diff
@@ -42,10 +57,16 @@ export function outputAsJson(
 
 function buildOutputObject(
     fileMetrics: Map<string, Map<string, MetricResult>>,
+    unknownFiles: string[],
     relationshipMetrics: CouplingResult
 ) {
-    const output: { nodes: OutputNode[]; relationships: OutputRelationship[] } = {
+    const output: {
+        nodes: OutputNode[];
+        info: OutputInfoNode[];
```
mylinhdao, line 65: replace "info" by "unknownFile" would make more sense to me

src/parser/GenericParser.test.ts
```diff
@@ -674,4 +675,33 @@ describe("GenericParser", () => {
...
+        it("should list files with unknown file extension", async () => {
+            const inputPath = fs.realpathSync(unknownTestResourcesPath);
+            const parser = new GenericParser(getParserConfiguration(inputPath));
+            const results = await parser.calculateMetrics();
+
+            const filePath = fs.realpathSync(unknownTestResourcesPath + "example.unknownExtension");
+            expect(results.unknownFiles.includes(filePath)).toBe(true);
+        });
+
+        it("should list files with no file extension", async () => {
```
mylinhdao, line 689: this seems to be the identical test as the test above

src/parser/helper/Helper.ts
```diff
@@ -90,19 +90,28 @@ export function formatPrintPath(filePath: string, config: Configuration): string
...
         if (fileExtension.length > 0) {
-            return { fileExtension: fileExtension, filePath: filePath };
+            let language = fileExtensionToLanguage.get(fileExtension);
+            if (language === undefined) {
+                language = Languages.Unknown;
+            }
+            return { fileExtension: fileExtension, filePath: filePath, language: language };
         }
     }
+    return { fileExtension: "", filePath: filePath, language: Languages.Unknown };
```
mylinhdao, line 114: you should return { fileExtension: fileExtension, ......

mylinhdao (commented):
beside of some naming conventions, its fine :)

src/parser/helper/Helper.ts
ResistantBear: No, thats correct if there is no file extension

src/parser/GenericParser.test.ts
ResistantBear: Fixed

mylinhdao (approved):
LGTM

src/commands/outputMetrics.ts
ResistantBear: clarified

src/commands/outputMetrics.ts
ResistantBear: clarified

PR #49, opened 2024-01-25
Bugfix for real lines of code with empty files

### LINKED ISSUE #47: Real lines of code is wrong for empty files
Real lines of code is 1 for empty files, but should be 0.

ResistantBear:
Empty files have now a rloc-value of 0 instead of 1, as root nodes without child nodes are no longer counted.

Resolves #47

### DIFF
--- a/src/parser/GenericParser.test.ts
+++ b/src/parser/GenericParser.test.ts
@@ -388,6 +388,16 @@ describe("GenericParser", () => {
             );
         });
 
+        it("should count correctly for a file with a single comment", async () => {
+            const inputPath = fs.realpathSync(tsTestResourcesPath + "single-comment.ts");
+            const parser = new GenericParser(getParserConfiguration(inputPath));
+            const results = await parser.calculateMetrics();
+
+            expect(results.fileMetrics.get(inputPath)?.get("real_lines_of_code")?.metricValue).toBe(
+                0
+            );
+        });
+
         it("should count correctly if there is a comment in the same line as actual code", async () => {
             const inputPath = fs.realpathSync(tsTestResourcesPath + "same-line-comment.ts");
             const parser = new GenericParser(getParserConfiguration(inputPath));
--- a/src/parser/GenericParser.test.ts
+++ b/src/parser/GenericParser.test.ts
@@ -191,6 +191,16 @@ describe("GenericParser", () => {
             );
         });
 
+        it("should count correctly for an empty file", async () => {
+            const inputPath = fs.realpathSync(phpTestResourcesPath + "empty.php");
+            const parser = new GenericParser(getParserConfiguration(inputPath));
+            const results = await parser.calculateMetrics();
+
+            expect(results.fileMetrics.get(inputPath)?.get("real_lines_of_code")?.metricValue).toBe(
+                0
+            );
+        });
+
         it("should count correctly if there is a comment in the same line as actual code", async () => {
             const inputPath = fs.realpathSync(phpTestResourcesPath + "same-line-comment.php");
             const parser = new GenericParser(getParserConfiguration(inputPath));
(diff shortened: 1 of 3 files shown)


src/parser/GenericParser.test.ts
```diff
@@ -191,6 +191,16 @@ describe("GenericParser", () => {
...
         });
 
+        it("should count correctly for an empty file", async () => {
+            const inputPath = fs.realpathSync(phpTestResourcesPath + "empty.php");
+            const parser = new GenericParser(getParserConfiguration(inputPath));
+            const results = await parser.calculateMetrics();
+
+            expect(results.fileMetrics.get(inputPath)?.get("real_lines_of_code")?.metricValue).toBe(
+                0
+            );
```
ce-bo, line 201: I just thought (because we have many such tests) that this could be shortened like this (pseudo code):
```suggestion
            const { fileMetrics } = await calculateMetrics(phpTestResourcesPath + "empty.php");

            expect(fileMetrics.get(inputPath)?.get("real_lines_of_code")?.metricValue).toBe(
                0
            );
```

Note: `calculateMetrics` would be a new method that creates the GenericParser and calls `realpathSync`

ce-bo: @mylinhdao this could be interesting to you, I guess :-)

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -56,14 +56,16 @@ export class RealLinesOfCode implements Metric {
...
-        cursor.gotoFirstChild();
-        const sureCodeLines = this.walkTree(cursor);
 
-        dlog("Included lines for rloc: ", sureCodeLines);
-
-        const rloc = sureCodeLines.size;
+        // Assume the root node is always some kind of program/file/compilation_unit stuff.
+        // So if there are no child nodes, the file is empty.
+        if (cursor.gotoFirstChild()) {
+            const sureCodeLines = this.walkTree(cursor);
```
ce-bo, line 65:
```suggestion
            const sourceCodeLines = this.walkTree(cursor);
```

ce-bo (approved):
LGTM but please fix one typo :-)

BridgeAR (commented):
What would happen with a file that would only contain e.g., a comment? I guess that would not be handled correct?

To overcome that issue, we could just skip the root node while adding the code lines in `walkTree`, right?

src/parser/GenericParser.test.ts
ResistantBear: Something similar to is already done by `testFileMetrics(...)` in #42. These new test cases will have to be refactored to also use `testFileMetrics` once merged. And put into the right file, too. Merge will be wonderful :D

src/parser/metrics/RealLinesOfCode.ts
ResistantBear: That was actually no typo. "sure code lines", because we can be sure there is code in them :laughing: 
But yeah, no one will understand that, I will change it.

ResistantBear:
> What would happen with a file that would only contain e.g., a comment? I guess that would not be handled correct?
> 
> To overcome that issue, we could just skip the root node while adding the code lines in `walkTree`, right?

No, that works just fine. The code already skips the root node (usually something like "file" or "compilation_unit") before calling `walkTree` by doing `gotoFirstChild` . The issue was that I had not checked if there **is** actually any node after the root node to skip to, which lead to `walkTree` being called with the root node in case the file is empty.

BridgeAR (approved):

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -31,39 +31,41 @@ export class RealLinesOfCode implements Metric {
      * {@link https://tree-sitter.github.io/tree-sitter/using-parsers#walking-trees-with-tree-cursors|Tree-sitter documentation},
      * this is the most efficient way to traverse a syntax tree.
      * @param cursor A {@link TreeCursor} for the syntax tree.
-     * @param sureCodeLines A set in which the line numbers of the found code lines are stored.
+     * @param realCodeLines A set in which the line numbers of the found code lines are stored.
```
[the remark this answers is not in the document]
ResistantBear: Ok

PR #57, opened 2024-01-31
Bugfix Java Comments #56

### LINKED ISSUE #56: Comment node does not work with Java, Update Snapshots
The "comment" node is not accepted as part of a query for java, despite it being a named node.
Update the snapshots to match the current state of the main.

ResistantBear:
Closes #56


mylinhdao (approved):
seems like comment node is not allowed to be queried for Java

PR #59, opened 2024-01-31
Fixes rloc for multiline strings in java

### LINKED ISSUE #58: Real lines of code fails with new grammar for java when using multi-line string literals
Code like the following does fail with the new Java grammar:

```Java
String multiLine = """
    Multi
    Line
   String
""";
```

ResistantBear:
Counts now for any lines over which a leaf node (= A single token of the language) spans to correctly handle multiline strings with the new Java grammar. In the new grammar, `(multiline_string_fragment)` is a node spanning over multiple lines without having any children.

This code assumes that the following example should count as 6 real lines of code. That means that we also count empty lines inside multiline strings as real lines of code.
```Java
String multiLine = """
    Multi

    Line
   String
""";
```

Closes #58

### DIFF
--- a/src/parser/metrics/RealLinesOfCode.ts
+++ b/src/parser/metrics/RealLinesOfCode.ts
@@ -39,8 +39,14 @@ export class RealLinesOfCode implements Metric {
             // Assume that first and last line of whatever kind of node this is, is a real code line.
             // This assumption should hold for all kinds of block/composed statements in (hopefully) all languages.
             realLinesOfCode.add(cursor.startPosition.row);
-            // Adding the last line is not necessary, as every last line has to have some syntactical element,
-            // which is again expressed as another syntax node.
+
+            // This is a leaf node, so add further lines if this single token of actual code
+            // spans over multiple lines and is no line ending:
+            if (cursor.currentNode.childCount == 0 && !"\r\n".includes(cursor.currentNode.type)) {
+                for (let i = cursor.startPosition.row + 1; i <= cursor.endPosition.row; i++) {
+                    realLinesOfCode.add(i);
+                }
+            }
         }
         // Recurse, depth-first
         if (cursor.gotoFirstChild()) {


src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -39,8 +39,14 @@ export class RealLinesOfCode implements Metric {
             // Assume that first and last line of whatever kind of node this is, is a real code line.
             // This assumption should hold for all kinds of block/composed statements in (hopefully) all languages.
             realLinesOfCode.add(cursor.startPosition.row);
-            // Adding the last line is not necessary, as every last line has to have some syntactical element,
-            // which is again expressed as another syntax node.
+
+            // This is a leaf node, so add further lines if this single token of actual code
+            // spans over multiple lines and is no line ending:
+            if (cursor.currentNode.childCount == 0 && !"\r\n".includes(cursor.currentNode.type)) {
```
BridgeAR, line 45: Are we certain it is always \r\n? Or could it also be e.g., a sole \n?

BridgeAR (approved):
LGTM

I believe we could change the set to a counter and use the last line as memory. We increase the line count as long as they node's start position is higher than the current last entry. That would reduce the memory overhead.

src/parser/metrics/RealLinesOfCode.ts
ResistantBear: As "includes" is used here, it actually should work with all three cases \n, \r\n and \r

ResistantBear:
> LGTM
> 
> I believe we could change the set to a counter and use the last line as memory. We increase the line count as long as they node's start position is higher than the current last entry. That would reduce the memory overhead.

I have tested that quickly but it lead to errors. I can image that it works after considering some special cases, but as this fix is needed to work on our test cases issue, I will go with the current solution and might look into further optimizations later. I have opened #63 for this.

PR #66, opened 2024-02-02
Document complexity metric and rename "mcc" to "complexity"

### LINKED ISSUE #41: Add definition of "McCabe

ResistantBear:
Defined the complexity-metric in the readme, similar to the "Complexity"-metric of SonarQube. This replaces the former "mcc" metric, as it did not equal the actual McCabe Cyclomatric Complexity.

For now, this metric does not include standard library functions like "foreach" or "map". If it is requested to also count calls of such functions, that should be discussed, implemented, tested and documented in a separate feature request.

Closes #41


README.md
```diff
@@ -43,23 +43,34 @@ This is quite slow and can take up to one or two hours but can provide good resu
...
     -   C++
-    -   Java
     -   JavaScript
     -   Kotlin
-    -   Python
 
-### Supported Metrics
+### Supported File Metrics
+
+**complexity**<br>
```
ce-bo, line 57: Let's mention in the PR that mcc was rename to complexity.

resources/Import-Grammars-Test/nodeTypesChanges_Reference.csv
```diff
@@ -566,16 +566,16 @@ Name:;Added to language(s):;Removed from language(s):;Remains in language(s):;Us
 "super";;kt;java,js,ts;;
```
ce-bo, line 566: I think it is not ideal to have a file like this which needs to be maintained. This information is available in other files already (nodeTypesConfig), isn't it?

resources/php/multiple-return-statements.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * McCabeComplexity test resource file
+ * Complexity test resource file
```
ce-bo, line 4: We could remove the commentse by moving the file into a "complexity" folder. By that, we would not need to touch exisitng files e.g. when a metric is renamed.

resources/Import-Grammars-Test/nodeTypesChanges_Reference.csv
ResistantBear: I left this when updating the import script to later add an automated test case for it. Additionally to the old nodeTypesConfig.json, we would also need fixed copies of the node-types.json of the grammars to get a reproducable result.

No, it is not the same as even our current nodeTypesConfig, as it does not include any of the manually added additions we did after importing the grammars.

resources/php/multiple-return-statements.php
ResistantBear: The comment is not needed at all, I guess. Most files do not include such a comment. This one did and setting the renaming tool of the IDE to change the name also in comments and strings has therefore also renamed that one.
Currently, some tests files are also used for multiple test cases. I found that quite useful to avoid redundancy and in the end, all metrics are (and should be able to be) calculated for each file. 

Thinking of it, we may restructure the test cases to make metric-gardener run once on all test files and then let the test cases run assertions on the results. That should speedup the tests runs and dramatically reduce the probability of the race condition to occur 🤔

ResistantBear: But that is not really related to this PR and could be done regardless of the folder structure.

README.md
ResistantBear: Done

ResistantBear:
Converted to draft until the metric definition is further clearified in the corresponding issue #41 
Also, there might be some test cases missing yet.

mylinhdao (approved):
LGTM

PR #69, opened 2024-02-02
Add tests for file metrics for C# #44

michaelalper-cezik-mw:
Added tests for C# similar to existing ones in TypeScript and PHP.


resources/c-sharp/real-lines-of-code.cs
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * A class description comment.
+ */
+public class Example {
+    /**
+     * This is a doc comment. Very good method.
+     * Indeed.
+     * @input Some input string.
+     */
+    exampleMethod(string input) {
```
ResistantBear: Missing return type

src/parser/config/nodeTypesConfig.json
```diff
@@ -337,7 +337,7 @@
     },
     {
         "expression": "is_pattern_expression",
-        "metrics": ["mcc"],
```
ResistantBear, line 340: This fixes that the code counted the pattern match (x is true) as a logical expression, but that should only be the case for and and or patterns.

ResistantBear (approved):
LGTM after our minor changes

resources/c-sharp/if-statements.cs
```diff
@@ -0,0 +1,36 @@
+int x, y;
+double z;
+
+if (x > y) {
```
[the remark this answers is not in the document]
ResistantBear: Opened issue #75 for this discussion

PR #73, opened 2024-02-05
Exclude Supertype Nodes

### LINKED ISSUE #61: Exclude node types that have subtypes in the import script
There was an issue with the updates Java-grammar, that the "comment" node type has now two new subtypes for "line_comment" and "block_comment" (see #56). Running a query for the "comment" node type resulted in a tree-sitter query exception. 
For now, the "comment" node type was manually deactivated for Java by using the "activated_for_languages"-field. However, if such "super node types" which have subtypes are generally something like an "abstract node type" which cannot be queried - and it looks like that is the case - the import script should not include these node types into the nodeTypesConfig.json (for that language), but only the subtypes.

ResistantBear:
This change makes the import script exclude all Supertype Nodes (node types that have subtypes) and uses said script to remove all Supernode Types from the current nodeTypesConfig.json.

As confirmed by the section about "Supertype Nodes" in the tree-sitter documentation, such types are abstract https://tree-sitter.github.io/tree-sitter/using-parsers
Therefore, a query for these node types causes an excpetion.

Closes #61

### DIFF
--- a/src/commands/import-grammars/ImportNodeTypes.ts
+++ b/src/commands/import-grammars/ImportNodeTypes.ts
@@ -163,10 +163,11 @@ async function updateLanguage(
             continue;
         }
 
-        toRemove.delete(grammarNodeType.type);
-        updateOrAddExpression(languageAbbr, grammarNodeType.type);
-
-        if (grammarNodeType.subtypes !== undefined) {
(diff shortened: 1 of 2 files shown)


mylinhdao (approved):
LGTM. I also don't see any problem with removing theses supertypes.

PR #74, opened 2024-02-05
Special treatment for python block comments

### LINKED ISSUE #31: Comment blocks in python are not handled

ResistantBear:
Adds special handling for python multiline comments, which are multiline strings that are neither stored in a variable/constant nor passed to a function.

Closes #31

### DIFF
--- a/src/parser/metrics/CommentLines.ts
+++ b/src/parser/metrics/CommentLines.ts
@@ -1,9 +1,14 @@
 import { QueryBuilder } from "../queries/QueryBuilder";
-import { ExpressionMetricMapping, QueryStatementInterface } from "../helper/Model";
+import {
+    ExpressionMetricMapping,
+    QueryStatementInterface,
+    SimpleQueryStatement,
+} from "../helper/Model";
 import { getQueryStatements } from "../helper/Helper";
 import { FileMetric, Metric, MetricResult, ParseFile } from "./Metric";
 import Parser, { QueryMatch, SyntaxNode } from "tree-sitter";
 import { debuglog, DebugLoggerFunction } from "node:util";
+import { Languages } from "../helper/Languages";
 
 let dlog: DebugLoggerFunction = debuglog("metric-gardener", (logger) => {
     dlog = logger;
@@ -25,6 +30,16 @@ export class CommentLines implements Metric {
     }
 
     async calculate(parseFile: ParseFile, tree: Parser.Tree): Promise<MetricResult> {
+        switch (parseFile.language) {
+            case Languages.Python:
+                this.statementsSuperSet.push(
+                    new SimpleQueryStatement(
+                        "(expression_statement (string)) @python_multiline_comment"
+                    )
+                );
+                break;
+        }
+
         const queryBuilder = new QueryBuilder(parseFile, tree);
         queryBuilder.setStatements(this.statementsSuperSet);
 
--- a/src/parser/metrics/RealLinesOfCode.ts
+++ b/src/parser/metrics/RealLinesOfCode.ts
@@ -71,13 +71,11 @@ export class RealLinesOfCode implements Metric {
     async calculate(parseFile: ParseFile, tree: Parser.Tree): Promise<MetricResult> {
         let isCommentFunction: (node: Parser.SyntaxNode) => boolean;
 
-        switch (parseFile.language) {
-            case Languages.Python:
-                isCommentFunction = (node: Parser.SyntaxNode) => this.isPythonComment(node);
-                break;
-            default:
-                isCommentFunction = (node: Parser.SyntaxNode) =>
-                    this.commentStatementsSet.has(node.type);
+        if (parseFile.language == Languages.Python) {
+            // Special treatment for python multiline comments:
+            isCommentFunction = (node: Parser.SyntaxNode) => this.isPythonComment(node);
+        } else {
+            isCommentFunction = (node: Parser.SyntaxNode) => this.isComment(node);
         }
 
         let rloc = 0;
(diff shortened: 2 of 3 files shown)


src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -32,10 +33,15 @@ export class RealLinesOfCode implements Metric {
      * this is the most efficient way to traverse a syntax tree.
      * @param cursor A {@link TreeCursor} for the syntax tree.
      * @param realLinesOfCode A set in which the line numbers of the found code lines are stored.
+     * @param isComment Function that checks whether a node is a comment or not.
      */
-    walkTree(cursor: TreeCursor, realLinesOfCode = new Set<number>()) {
+    walkTree(
```
mylinhdao, line 38: i suggest to not pass a lambda function as parameter to make the code easy to understand. Let's write a function to prove if the syntax node is a comment in a specific language. That approach may lead to minimal increased runtime, but it spares time in development for later.

mylinhdao (commented):
LGTM. Pls consider my suggestion

src/parser/metrics/CommentLines.ts
```diff
@@ -25,6 +30,16 @@ export class CommentLines implements Metric {
...
     async calculate(parseFile: ParseFile, tree: Parser.Tree): Promise<MetricResult> {
+        switch (parseFile.language) {
+            case Languages.Python:
+                this.statementsSuperSet.push(
+                    new SimpleQueryStatement(
+                        "(expression_statement (string)) @python_multiline_comment"
+                    )
+                );
+                break;
+        }
```
ce-bo, line 41: Let's don's use switches as long as we don't actually need them. An if is sufficient here.

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -62,13 +69,24 @@ export class RealLinesOfCode implements Metric {
...
+        let isCommentFunction: (node: Parser.SyntaxNode) => boolean;
+
+        switch (parseFile.language) {
+            case Languages.Python:
+                isCommentFunction = (node: Parser.SyntaxNode) => this.isPythonComment(node);
+                break;
+            default:
+                isCommentFunction = (node: Parser.SyntaxNode) =>
+                    this.commentStatementsSet.has(node.type);
+        }
```
ce-bo, line 81:
```suggestion
        let isCommentFunction = parseFile.language === Languages.Python ? this.isPythonComment(node) : this.commentStatementsSet.has(node.type);
```

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -81,6 +99,20 @@ export class RealLinesOfCode implements Metric {
...
+     */
+    isPythonComment(node: Parser.SyntaxNode) {
+        return (
+            this.commentStatementsSet.has(node.type) ||
+            (node.type === "expression_statement" &&
+                node.childCount === 1 &&
+                node.child(0)?.type === "string")
+        );
+    }
+
```
ce-bo, line 115: I think the comment does not add value. So let's omit if to have a better maintainability.
```suggestion
    isPythonComment(node: Parser.SyntaxNode) {
        return
            this.commentStatementsSet.has(node.type) ||
            (node.type === "expression_statement" &&
                node.childCount === 1 &&
                node.child(0)?.type === "string");
    }
```

ResistantBear: I think we should mention what the special treatment for Python exactly does. But it does not have to be a fully-fledged JSDoc-Comment, I agree.

ce-bo: Hm okay. Then we should maybe do it like this and name the function for what it is actually doing. By that, I think a comment is not needed anymore. Maybe a comment is needed at the place where this method is called.
```suggestion
    isPythonMultiLineComment(node: Parser.SyntaxNode) {
        return (
            this.commentStatementsSet.has(node.type) ||
            (node.type === "expression_statement" &&
                node.childCount === 1 &&
                node.child(0)?.type === "string")
        );
    }

```

ResistantBear: Well, it does also check whether it is an ordinary comment (`this.commentStatementsSet.has(node.type)`) as python does also include those, so this name would be misleading. Maybe the comment was misleading as well in that case.

ResistantBear: I replaced the JSDoc-comment with a more clear documentation of what the special handling actually does and added a comment to the usage side as well.

src/parser/metrics/RealLinesOfCode.ts
ResistantBear: I tend to disagree, as having the distinction of cases for the programming languages at the beginning of the calculate()-function makes it more obvious that there is a special treatment for a language here and where to add a new special treatment later. Also I don't like making things less efficient, yes. And it should not be too hard to understand what is going on here.

src/parser/metrics/CommentLines.ts
ResistantBear: Replaced the switch-case with an if-condition

src/parser/metrics/RealLinesOfCode.ts
ResistantBear: I have now done the following, as "isPythonSingleLineComment" is identical to how comments are detected in all other languages:

```typescript
isPythonComment(node: Parser.SyntaxNode) {
        return this.isComment(node) || this.isPythonMultilineComment(node);
}
```

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -81,6 +96,24 @@ export class RealLinesOfCode implements Metric {
         };
     }
 
+    isComment(node: Parser.SyntaxNode) {
+        return this.commentStatementsSet.has(node.type);
+    }
+
+    isPythonComment(node: Parser.SyntaxNode) {
```
mylinhdao, line 103: beautiful :)

mylinhdao (approved):
LGTM

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -81,6 +96,24 @@ export class RealLinesOfCode implements Metric {
...
+    }
+
+    isPythonMultilineComment(node: Parser.SyntaxNode) {
+        // Multiline comments in python are (multiline) strings that are
+        // neither assigned to a variable nor used as call parameter.
+        return (
+            node.type === "expression_statement" &&
+            node.childCount === 1 &&
+            node.child(0)?.type === "string"
+        );
```
ce-bo, line 114: parenthesis are not needed :-)

ResistantBear: Oh, I was sure I removed them. How embarrassing. 
As My Linh already merged this, I have to open a new PR for this.

Issue #85, opened 2024-02-06
Add complexity and function tests for Java language


mylinhdao:
After the complexity metric is defined in PR #41 , we need new tests for Java.

PR #86, opened 2024-02-07
Add ci pipeline to project

### LINKED ISSUE #62: Setup CI pipeline to execute unit tests and other QA tasks
Let's check what we do in [CodeCharta ](https://github.com/MaibornWolff/codecharta/tree/main/.github/workflows) and adapt the CI Pipeline for metric-gardener.
In the first step, tests and prettier should be executed.

mylinhdao:
Close #62

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,3 +3,5 @@
 resources
 .*
 dist
+__snapshots__
+LICENSE
--- a/resources/java/functions/LambdaExpression.java
+++ b/resources/java/functions/LambdaExpression.java
@@ -1,0 +1,16 @@
+// Functional interface with a single abstract method (SAM)
+interface MyFunctionalInterface {
+    void myMethod(String s);
+    void myMethod2();
+}
+
+public class LambdaExample {
+    public static void main(String[] args) {
+        // Using a lambda expression to implement the abstract method of the functional interface
+        MyFunctionalInterface myLambda = (String s) -> System.out.println("Hello, " + s);
+        MyFunctionalInterface myLambda2 = () -> System.out.println("Hello");
+
+        // Calling the method defined in the lambda expression
+        myLambda.myMethod("World");
+    }
+}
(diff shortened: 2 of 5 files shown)


resources/java/functions/LambdaExpression.java
```diff
@@ -0,0 +1,16 @@
+// Functional interface with a single abstract method (SAM)
```
ce-bo, line 1: Did you add this file in this PR? I think this should not be part of this PR, right?

.prettierignore
```diff
@@ -2,4 +2,6 @@
 !src/*
 resources
 .*
-dist
\ No newline at end of file
+dist
+__snapshots__
+LICENSE
```
ce-bo, line 7: The Licence file should not be ignored

ce-bo (approved):
In general LGTM but please check my remarks please.

.prettierignore
michaelalper-cezik-mw: But during format:check we get this error otherwise:
`Error: No parser could be inferred for file: LICENSE`
The job still succeeds but it seems unnecessary to check the format of the license file.

ce-bo: If there is no parser for the format check, this should not lead us to ignore the file. 
Let's rather think of ignoring the file in the config of the format checker. Another way would be just to change the LICENCE file to LICENCE.md. I think I forgot to add the .md prefix when I created the file. :-)

ResistantBear: Isn't the `.prettierignore`-file not already the config file of the formatter for making it ignore certain files?

resources/java/functions/LambdaExpression.java
ResistantBear: This looks like a file for issue #85, indeed. It seems like this has been removed from this PR, good.

.prettierignore
ResistantBear: I note that the LICENCE-file was renamed to be a .md-file. We probably could also ignore all files without file extension. But its fine for now.

ResistantBear:
> In general LGTM but please check my remarks please. However I would execute the pipeline only for PRs or pushes to main to save resources.

Seems like this has been reconfigured to your suggestion. So I go ahead to merge this now.

PR #91, opened 2024-02-07
Fix: remove side effect that caused race condition in comment lines

### LINKED ISSUE #90: Python comment_lines metric has a race condition
The special handling for python multiline comments for the comment lines metric does not work correctly if more than one file is being processed in one run of the metric-gardener.

ResistantBear:
Removed a side effect on `this.commentStatementSet` that lead to the multiple addition of the query used for finding python multiline comments when more than one python file is processed by metric-gardener.

Fixes #90


mylinhdao (approved):
LGTM

PR #89, opened 2024-02-07
Add test cases for the file metric calculation for C++

ResistantBear:
Adds test source and header files in the C++ programming languages. These files are tested to compile correctly using the LLVM C++ compiler clang that comes with MacOS (command `clang -c -xc++ -std=c++20 <file>`)
Adds test cases that check if the file metrics are calculated correctly if these C++ files are used as input. 

Additional notes:
- Fixes that C++ for-each-loops (syntax node `for_range_loop`) were not counted when calculating the complexity metric

Resolves #84

PR #92, opened 2024-02-07
Refactor test cases to reduce the times the parser is re-created

### LINKED ISSUE #76: May refactor test cases to run the parser only once per .test.ts file
Thinking of it, we may restructure the test cases inside one .test.ts file to make metric-gardener run once on all test files for that language and then let the test cases run assertions on these results. That should speedup the tests runs and dramatically reduce the probability of the race condition to occur 🤔

_Adopted from a post by @ResistantBear in https://github.com/MaibornWolff/metric-gardener/pull/66#discussion_r1477792077_

ResistantBear:
Should reduce the probability of the race condition ~~and improve the performance of the test suites.~~
Also helped to identify a bug that only occurs when more than one source code file is parsed in one run: #90 

Is based on #89, so do not merge this until #89 has been merged.

Resolves #76

### DIFF
--- a/test/parser/CPlusPlusMetrics.test.ts
+++ b/test/parser/CPlusPlusMetrics.test.ts
@@ -1,68 +1,69 @@
-import { testFileMetric } from "./TestHelper";
-import { FileMetric } from "../../src/parser/metrics/Metric";
+import { expectFileMetric, parseAllFileMetrics } from "./TestHelper";
+import { FileMetric, MetricResult } from "../../src/parser/metrics/Metric";
 
 describe("C++ metrics tests", () => {
     const cppTestResourcesPath = "./resources/c++/";
 
+    let results: Map<string, Map<string, MetricResult>>;
+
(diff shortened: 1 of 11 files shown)


ResistantBear:
Does not seem to achive a measureable performance benefit, so until we need this to avoid the race condition, the benefits of this are only that we can detect bugs like the one with python, because we currently always parse only a single file and never automatically test the case of parsing more than one file.

ResistantBear:
> We should have tests that do parse more than one file, as you said. How to continue here? @ResistantBear

In this case I can go ahead with this PR as it should also reduce the probability of the race condition to occur.

ResistantBear:
I am not sure about this. It does run minimaly slower than before and makes it harder to debug things, as running one test case only does also invoke parsing all files.
On the other hand, we should also test parsing more than one file to check for side effect bugs like #90.

michaelalper-cezik-mw (approved):
Looks good

Issue #64, opened 2024-02-02
Do not count for default cases in Java, Python, Kotlin and C++ switch case blocks


ResistantBear:
Currently, the code does not differentiate between case and default labels in a switch case statement in Java, while it does do so for C#. This is caused by the grammar for java using the same node_type for both labels.
Probably needs a special treatment in our code.

Once this is implemented, un-skip the test case for java switch case statements inside `JavaMetric.test.ts`.

ce-bo:
Thank you @ResistantBear. We should open a feature request for the java grammar to separate case and default labels in a switch statement.
For now it would be a pragmatic solution to just let it as it is which means to also count the default label even if this is not correct. It most probably does not falsify the results in a crucial way.

A language specific handling could be implemented, if there is no response from the dev-team of tree-sitter-java grammar after some time. I added priority medium for this, because there are other more important tasks for now anyway.

ResistantBear:
We have opened https://github.com/tree-sitter/tree-sitter-java/issues/171 to request this differentiation.

michaelalper-cezik-mw:
Same problem for Python, probably because default keyword is not used there. Instead, it is just another case with a underscore as wildcard like this:
`case _:`

Can be identified by a case_pattern node without any child nodes.

michaelalper-cezik-mw:
Same behaviour for Kotlin where the default branch is specified by `else -> ...`

The node type given by tree-sitter is the same as for the non-default branches (when_entry) but it can be identified by a missing when_condition child node.

ResistantBear:
Mh, that seems to get a common problem for most grammars, that they do not differentiate between case and default labels in try-catch-blocks, unlike C#. How unfortunate. 

The easiest solution would be to also count default-labels, then. That could be done by just changing the config-file. On the other hand, doing so would be not in line with the definition of the Cyclomatic Complexity as it does not lead to a new edge in the control flow, just like an else-statement, which we do not count either.

Having special treatments for various languages, or making the definition/calculation of that metric inconsistent. That's a difficult one.

ResistantBear:
I noticed that there is a test missing for switch-case-statements in C++. 
Unfortunately, the tree-sitter grammar for that language suffers from the same issue.

ResistantBear:
> Let's activate this test and asume that default labels are counted as well.
We agreed that it is ok for now to also count them. So let's relfect that in the tests but we should mention that in the metric definitions.

[...]

PR #93, opened 2024-02-08
Add tests for file metrics for Python

michaelalper-cezik-mw:
Adds test cases and source files and for Python programming language. Source files were tested to compile correctly.

Test for complexity metric of case-statements is set to skip because tree-sitter lacks node for default case, see #64 .


test/parser/PythonMetrics.test.ts
```diff
@@ -5,19 +5,117 @@ describe("Python metrics test", () => {
...
+
+        it("should not count multiple return statements within functions and methods correctly", async () => {
+            await testFileMetrics(
+                pythonTestResourcesPath + "multiple-return-statements.py",
+                FileMetric.complexity,
+                3
+            );
+        });
+
+        it("should count try-catch-finally properly by only counting the catch-block", async () => {
```
mylinhdao, line 48: can Python have more than 1 catch-block per try-block ? If yes, adding one more catch-block makes this test more precise I think.

mylinhdao (approved):
LGTM

PR #94, opened 2024-02-08
Add complexity and function metric tests for Java 

### LINKED ISSUE #85: Add complexity and function tests for Java language

mylinhdao:
~~But skip the test with method references, since the handling for them is not clear yet.~~
Closes #85


ResistantBear:
Method references seem to be the Java-concept for passing a function by value. As that does not involve implementing a new function (at least not from the view of the programmer), we should not count them as new function definition.

resources/java/functions/LambdaExpression.java
```diff
@@ -0,0 +1,21 @@
...
+    }
+    public static void main(String[] args) {
+        executeOperation(4, (x) -> x * 2); // Lambda expression for doubling the value
+
+        int squaredResult = square(3);
+        System.out.println("Square Result: " + squaredResult);
+    }
+    @FunctionalInterface
+    interface MyFunctionalInterface {
+        int performOperation(int value);
```
ResistantBear, line 19: That also counts as function definition here, mh... 
For C++, we (still) do not count function declarations without actual implementation at that place as a function currently and it definitely makes sense there. So it would again be a deviation between the metric result between different programming languages to count such pure function declarations/interfaces here. But there does not seem to be a way to change this behaviour for Java with the config file only, again :(

ResistantBear (approved):
LGTM, but check that thing with the for-loop in the if-statements file.

resources/java/complexity/IfStatements.java
```diff
@@ -22,5 +22,8 @@ public void exampleIfStatement(int value) {
         } else {
             System.out.println(value + " is equal to 0");
         }
+        for (;;) {
+            // Code for an infinite loop
+        }
```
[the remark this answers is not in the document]
mylinhdao: you are right, i just corrected that :)

resources/java/functions/LambdaExpression.java
ResistantBear: I had opened issue #98 to discuss that. Based on your suggestion, I guess that we should change the behaviour for C++ instead within the scope of that new issue.

ResistantBear (approved):

PR #99, opened 2024-02-09
Add support and tests for records and compact constructor #70

### LINKED ISSUE #70: Add support for Java and C# `record_declaration` and `compact_construct_declaration`
These are new features that are present in the currently used grammars, but not covered by our metrics / metric mappings now.
They should probably be treated similarly to how we already handle structs in other languages.

mylinhdao:
Counts records in Java and C# as classes, as they are a special form of class with more build-in features.

I removed the skip label for the switch statement test. We decided that default label should also be counted for complexity metric temporarily. This will also be mentioned in README.md.  
Closes #70

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,12 +61,11 @@ Counts expressions that branch the control flow (if-statements, loops, catch-blo
 -   function declarations (including lambda expressions)
 -   binary logical operations (like AND and OR)
 
-Note: In Java, the default label in a switch statement is counted for complexity metric.
 **functions**<br>
 The number of function definitions inside a file. Includes all kinds of functions, like constructors, lambda functions, member functions, etc.
 
 **classes**<br>
-The number of class definitions inside a file, also counting for enums and interfaces.
+The number of class definitions inside a file, also counting for enums, interfaces, structs and records.
 
 **lines_of_code**<br>
 The total number of lines of a file, including empty lines, comments, etc.
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Counts expressions that branch the control flow (if-statements, loops, catch-blo
 -   function declarations (including lambda expressions)
 -   binary logical operations (like AND and OR)
 
+Note: In Java, the default label in a switch statement is counted for complexity metric.
 **functions**<br>
 The number of function definitions inside a file. Includes all kinds of functions, like constructors, lambda functions, member functions, etc.
 
--- a/resources/c-sharp/records.cs
+++ b/resources/c-sharp/records.cs
@@ -11,3 +11,4 @@ public record struct Circle
     public double Y { get; init; }
 
 }
+public record class Ocean(string Address);
--- a/resources/c-sharp/records.cs
+++ b/resources/c-sharp/records.cs
@@ -1,0 +1,13 @@
+public record House(string Address, string City);
+public record Person
+{
+    public required string FirstName { get; init; }
+    public required string LastName { get; init; }
+};
+public readonly record struct Point(double X, double Y, double Z);
+public record struct Circle
+{
+    public double X { get; init; }
+    public double Y { get; init; }
+
+}
--- a/resources/java/classes/Records.java
+++ b/resources/java/classes/Records.java
@@ -1,0 +1,13 @@
+public record Person(String name, String address) {
+    public Person(String name) {
+        this(name, "Unknown");
+    }
+}
+public record Mensch (String name, String address) {
+    public Person {
+        Objects.requireNonNull(name);
+        Objects.requireNonNull(address);
+    }
+}
+public record House(String address) {
+}
--- a/resources/java/functions/LambdaExpression.java
+++ b/resources/java/functions/LambdaExpression.java
@@ -18,4 +18,8 @@ public class SampleFunctions {
     interface MyFunctionalInterface {
         int performOperation(int value);
     }
+<<<<<<< HEAD
 }
+=======
+}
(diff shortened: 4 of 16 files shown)


resources/java/classes/Records.java
```diff
@@ -0,0 +1,13 @@
+public record Person(String name, String address) {
+    public Person(String name) {
+        this(name, "Unknown");
+    }
+}
+public record Mensch (String name, String address) {
+    public Person {
```
ResistantBear, line 7: Have you tried to compile that code? I guess that it should not compile to have a compact constructor for "Person" in a record "Mensch"? 

You may also change the name "Mensch" to an english term like "human" or "employee" or whatever english word you like :)

ResistantBear (commented):
I agree with couting records as classes. However, you should mention that decision you made in the Pull Request. 

We should also document that decision in the Readme. I took a look on it, and the definition of the classes metric also does not mention that we count for structs, like they are in C++ and C#, so that should be added, too.

Take also a look at the other remarks. Other than that, it looks good.

test/parser/JavaMetrics.test.ts
```diff
@@ -50,6 +50,13 @@ describe("Java metrics tests.", () => {
                 0
             );
         });
+        it("should count all records.cs as classes", async () => {
```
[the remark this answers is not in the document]
mylinhdao: yes it was a mistake caused by renaming. I corrected that.

resources/c-sharp/records.cs
```diff
@@ -0,0 +1,13 @@
+public record House(string Address, string City);
+public record Person
+{
+    public required string FirstName { get; init; }
+    public required string LastName { get; init; }
+};
```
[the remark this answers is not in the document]
mylinhdao: i added `record class` as suggested

resources/java/classes/Records.java
mylinhdao: you are right, it could not be compiled. I have checked all java test resources on compile-ability and adjusted them afterwards.

test/parser/JavaMetrics.test.ts
```diff
@@ -218,7 +226,7 @@ describe("Java metrics tests.", () => {
             await testFileMetric(
                 javaTestResourcesPath + "/" + FileMetric.complexity + "/SwitchStatement.java",
                 FileMetric.complexity,
-                13
+                8
             );
         });
```
ResistantBear, line 230: Nit: add line break here for consistent style, also for the following "it" calls.

ce-bo (approved):
LGTM :-)

test/parser/JavaMetrics.test.ts
```diff
@@ -50,6 +50,13 @@ describe("Java metrics tests.", () => {
                 0
             );
         });
```
ResistantBear, line 52: Nit: add line break here. Also on the other cases where there is no line break between it (...) calls.

ResistantBear (commented):
Still no documentation about our decision to count structs and records for the classes metric in the README.md and the pull request description. Some minor remarks.

test/parser/JavaMetrics.test.ts
```diff
@@ -211,7 +225,7 @@ describe("Java metrics tests.", () => {
             await testFileMetric(
```
ResistantBear, line 225: The following also applies here:

Let's activate this test and asume that default labels are counted as well.
We agreed that it is ok for now to also count them. So let's relfect that in the tests

_Originally posted by @ce-bo in https://github.com/MaibornWolff/metric-gardener/pull/101#discussion_r1485851071_

ce-bo: Since there is a special requirement to count also default statements for now, let's have an individual test for exactly that. Otherwise too much requirements are covered here in the test which makes it hard to understand the implemented behavior.

README.md
```diff
@@ -61,6 +61,7 @@ Counts expressions that branch the control flow (if-statements, loops, catch-blo
 -   function declarations (including lambda expressions)
 -   binary logical operations (like AND and OR)
 
+Note: In Java, the default label in a switch statement is counted for complexity metric.
```
ResistantBear, line 64: Not only for Java, by I have added the corresponding documentation already to #101, so I might solve this when solving the merge conflict that will inevitably occur here now.

There is still no documentation that we count structs and records with the classes metric, however.

resources/java/functions/LambdaExpression.java
```diff
@@ -18,4 +18,8 @@ public static void main(String[] args) {
     interface MyFunctionalInterface {
         int performOperation(int value);
     }
+<<<<<<< HEAD
```
ResistantBear, line 21: Merge conflict

README.md
ResistantBear, line 64: There is still no documentation that we also count structs and records with the classes metric...

ResistantBear (changes requested):
Please resolve the merge conflict within `resources/java/functions/LambdaExpression.java` and add documentation for the classes metric to reflect that we also count structs and records now.

test/parser/JavaMetrics.test.ts
ResistantBear: It is now more defined as an "exception" from the normal behaviour that is currently allowed for that language rather than a special requirement that they are counted. 
My Linh has reduced the scope of this test case, I am ok with that now.

ResistantBear (approved):
All issues resolved

PR #105, opened 2024-02-12
Add tests for file metrics for Kotlin

michaelalper-cezik-mw:
Adds test cases and source files and for Kotlin programming language. Source files were tested to compile correctly.


test/parser/KotlinMetrics.test.ts
```diff
@@ -0,0 +1,170 @@
...
+
+        it("should count case but no default statements correctly", async () => {
+            await testFileMetric(
+                kotlinTestResourcesPath + "case-statements.kt",
+                FileMetric.complexity,
+                5
+            );
+        });
+
+        it("should count functions and methods correctly", async () => {
```
mylinhdao, line 32: Looks like init block is counted for complexity and functions metric for Kotlin, but not Java. Maybe we should mention this in our README.md? @ResistantBear @michaelalper-cezik-mw 
```suggestion
        it("should count all function declarations and one init block correctly", async () => {
```

test/parser/KotlinMetrics.test.ts
```diff
@@ -0,0 +1,170 @@
...
+            await testFileMetric(
+                kotlinTestResourcesPath + "functions-and-methods.kt",
+                FileMetric.complexity,
+                11
+            );
+        });
+
+        it("should not count multiple return statements within functions and methods correctly", async () => {
```
mylinhdao, line 40:
```suggestion
        it("should not count multiple return statements within functions and methods", async () => {
```

mylinhdao (commented):
In general LGTM, but the remarks must be discussed.

test/parser/KotlinMetrics.test.ts
```diff
@@ -0,0 +1,170 @@
...
+                FileMetric.complexity,
+                10
+            );
+        });
+
+        it("should not count any class declaration", async () => {
+            await testFileMetric(kotlinTestResourcesPath + "classes.kt", FileMetric.complexity, 4);
+        });
+
+        it("should count case but no default statements correctly", async () => {
```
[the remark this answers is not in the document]
michaelalper-cezik-mw: @ce-bo suggested we should not skip any tests (can't find the post).
After #64 is implemented these tests will fail and we can correct the expected complexity.

test/parser/KotlinMetrics.test.ts
ResistantBear: I guess the init-block is a block where you can throw arbitrary statements in, just like a function body? In this case, we should count it as a function, just like contructors, etc.

test/parser/KotlinMetrics.test.ts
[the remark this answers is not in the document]
ce-bo: To wait for #64 blocks this PR from being merged. I think you mean the same: I suggest to change the test name to also count default labels for now, as this is the current behavior of metric-gardener and update it in #64 later when the test fails.

test/parser/KotlinMetrics.test.ts
ce-bo: I agree that this is like a function and should be counted accordingly for complexity and functions metric.
Do we handle init blocks for java?

ce-bo:
I am super happy that you add such extensive tests with example files 🥳

test/parser/KotlinMetrics.test.ts
mylinhdao: @ce-bo actually, init block are not counted in Java. But we will change that in #108

mylinhdao (approved):
all remarks are solved

PR #107, opened 2024-02-12
Differentiate case labels from default labels for all supported languages

### LINKED ISSUE #64: Do not count for default cases in Java, Python, Kotlin and C++ switch case blocks

ResistantBear:
Differentiate case labels from default labels for languages where the grammar does not provide an own node type for default labels by checking for child nodes. Works with nodes that have a single child if they represent a case-label and no default label (fallbacks to this behaviour if new case label syntax nodes are encountered), as well as the two node types `case_statement` and `when_entry`, which have special child nodes if they represent a case-label.

Adds two new node Type categories „case_label“ and „default_label“. Applies the special handling to all case_label nodes if there is no default_label node type for a language. 

Refactors query statements using inheritance to be less redundand and moved them to an own file.

Resolves #64


src/parser/metrics/Complexity.ts
```diff
@@ -44,12 +44,94 @@ export class Complexity implements Metric {
...
+                    // checking if there is a child with the field name "value":
+                    this.complexityStatementsSuperSet.push(
+                        new SimpleLanguageSpecificQueryStatement(
+                            "(case_statement value: _ ) @case_statement",
+                            haveNoDefaultNodeType,
+                            caseNodeType.activated_for_languages
+                        )
+                    );
+                } else {
+                    // Check if the case label node has a named child:
```
ce-bo, line 119: That means if there is a named child, it is supposed to be a case label that should be counted, otherwise it could be a default one that should be ignored?

src/parser/metrics/Complexity.ts
```diff
@@ -44,12 +44,94 @@ export class Complexity implements Metric {
                     }
                 }
             }
-        });
+        }
+        this.createCaseQueryStatements(allNodeTypes);
+    }
+
+    createCaseQueryStatements(allNodeTypes: ExpressionMetricMapping[]) {
```
ce-bo, line 51: I recommend to refactor this function to make it more readable. It's complexity is roughly 12 which is a little bit too high.

src/parser/metrics/Complexity.ts
ResistantBear: Yes. That works e.g. for Java.

src/parser/metrics/Complexity.ts
ResistantBear: I have already splitted out a part of this now, as the length of this function further escalated as I added a special treatment for Kotlins `when_entry`

ResistantBear: Ah, ok, you meant the constructor. I will check that.

ResistantBear: Refactored the constructor and saved one loop over all node types, which should also improve the performance.

michaelalper-cezik-mw (approved):
Sieht sehr gut aus, elegante Lösung des Problems und sinnvolles Refactoring.

src/parser/metrics/Complexity.ts
```diff
@@ -1,55 +1,165 @@
...
+            // if the first child ist a "when_condition":
+            this.complexityStatementsSuperSet.push(
+                new SimpleLanguageSpecificQueryStatement(
+                    "(when_entry (when_condition)) @when_entry",
+                    noDefaultLangAbbrs,
+                    caseDefaultNodeType.activated_for_languages
+                )
+            );
+        } else {
+            // Standard treatment: check if the case label node has a named child:
```
ce-bo, line 146: It is hard to see that you try to identify default labels in this else block. I am not super happy to have so much comments, but we should at least extend the comment to make that clear.

ce-bo:
Thank you for the implementation. I fear that we should be careful with upcoming feature requests like this to support issues in grammars of certain languages. It adds not too much value in this case, so that it is worth in all cases to cover those cases with metric-gardener. But that's just a personal learning :-)

mylinhdao (approved):
The code is quite readable, which is commendable. However, the computation of query statements occurs each time the program is run, resulting in unnecessary resource consumption. Notably, the result remains constant unless changes are made to 'nodeTypeConfig.json.' To optimize this process, consider relocating the logic to 'import-grammars' and storing the query statements in a file. This adjustment would alleviate the need for frequent recalculations, enhancing overall efficiency. 
We can merge this since the logic is correct and the code is good :)

ResistantBear:
> Thank you for the implementation. I fear that we should be careful with upcoming feature requests like this to support issues in grammars of certain languages. It adds not too much value in this case, so that it is worth in all cases to cover those cases with metric-gardener. But that's just a personal learning :-)

Yes, we should take care that the number of special treatments does not escalate as that would be cumbersome to maintain when installing new grammars. I tried to make this as generic as possible to also handle if similar nodes appearing in another grammar while on the same time handle the case that there are node types in a grammar that has an explicit default label with the same name as the special-handled ones. Nevertheless, it is likely that other grammars again introduce new nodes for case-labels.

ResistantBear:
> The code is quite readable, which is commendable. However, the computation of query statements occurs each time the program is run, resulting in unnecessary resource consumption. Notably, the result remains constant unless changes are made to 'nodeTypeConfig.json.' To optimize this process, consider relocating the logic to 'import-grammars' and storing the query statements in a file. This adjustment would alleviate the need for frequent recalculations, enhancing overall efficiency. We can merge this since the logic is correct and the code is good :)

You have already opened #113 to disuss the optimization. As that issue also applies to all other file metrics, it is out of scope of this PR.

src/parser/metrics/Complexity.ts
```diff
@@ -1,55 +1,165 @@
...
-                    } else {
-                        this.mccStatementsSuperSet.push(
-                            new ExpressionQueryStatement(
-                                expression,
-                                languages,
-                                activated_for_languages
-                            )
-                        );
-                    }
+        const withDefaultLabelNodeType = new Set<string>();
```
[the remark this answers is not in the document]
ResistantBear: The abbreviations of these languages, to be precise. So it should be named `abbreviationsOfLanguagesWithDefaultLabelNodeType` but I considered that as too long :D

src/parser/metrics/Complexity.ts
```diff
@@ -1,55 +1,165 @@
...
+    /**
+     * Add queries for each case label syntax node type.
+     */
+    addCaseLabelQueryStatements(
+        caseNodeTypes: ExpressionMetricMapping[],
+        languagesWithDefaultLabelAbbr: Set<string>
+    ) {
+        for (const caseNodeType of caseNodeTypes) {
+            const haveDefaultNodeType: string[] = [];
+            const haveNoDefaultNodeType: string[] = [];
```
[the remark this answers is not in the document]
ResistantBear: `haveDefaultNodeType` is actually different to the call parameter `languagesWithDefaultLabelAbbr`, as these are the abbreviations of the languages which have the current `caseNodeType` that also have a default node type. 
`haveNoDefaultNodeType` are the abbreviations of the languages which have the current `caseNodeType` that do not have a default node type.

I had actually no good idea how to include this information in the name of the variable, so if you have a good recommendation, I will listen.

src/parser/metrics/Complexity.ts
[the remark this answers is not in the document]
ResistantBear: I will rename it to `languagesWithDefaultLabelAbbr` like the call parameter of `addCaseLabelQueryStatements`

src/parser/helper/QueryStatements.ts
```diff
@@ -0,0 +1,117 @@
...
+    constructor(
+        expression: string,
+        applicable_for_languages: string[],
+        activated_for_languages?: string[]
+    ) {
+        super(applicable_for_languages, activated_for_languages);
+        this.expression = expression;
+    }
+
+    override toQuery(): string {
```
BridgeAR, line 75: Let's stay in the JS world and use `toString()` instead of `toQuery()`. `toString()` is always called, when serializing an object, e.g., while doing: `object + ''`.

src/parser/helper/QueryStatements.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    applicableFor(language: Languages): boolean {
+        return this.applicableForLanguages.has(language);
+    }
+
+    abstract toQuery(): string;
+}
+
+export class ExpressionQueryStatement extends LanguageSpecificQueryStatement {
+    private readonly expression: string;
```
BridgeAR, line 64: I suggest to use `#expression` instead of `private expression`. It's the way in JS to make a variable actually private. The keyword private is used in TS but it's just a fake private thing and it's still accessible from the outside. I believe we also do not have to mark it as readonly, when using a private variable.

src/parser/metrics/Complexity.ts
ResistantBear: The comments primarily serve the purpose to document what the speciality of the syntax node type is that is handled here, they are not really meant to explain how the code works.

However, I admit that I may overdo when it comes to wordy documentation. In this case, however, the others told me that they liked the comments here, too, so I suggest to keep them.
But I have nothing against making my comments even longer :D

src/parser/helper/QueryStatements.ts
ResistantBear: Ok. We only use it in QueryBuilder.ts, but now that I already refactor that code, I might as well change that.

src/parser/helper/QueryStatements.ts
ResistantBear: Ah, good to know, I will change all private fields to # then.
That `readonly` thing is a recommendation from the IDE that still pops up when using #. I guess that making that field readonly to also prevent modification from inside the class does not do any harm.

mylinhdao (approved):
LGTM

Issue #65, opened 2024-02-02
Race condition when running all tests files


ResistantBear:
We have encountered an issue where some tests "just fail" because of tree.rootNode being undefined when running all tests, e.g. via "npm test". This seems to be triggered by the higher number of test files or test cases in this PR, but also happened seldomly before. 

As it turns out, this is probably caused by a race condition in the tree-sitter library.
See https://github.com/tree-sitter/node-tree-sitter/issues/181
There is a potential fix for this linked, but it never got merged:
https://github.com/tree-sitter/node-tree-sitter/pull/75

_Originally posted by @ResistantBear in https://github.com/MaibornWolff/metric-gardener/issues/42#issuecomment-1918743774_

ResistantBear:
Unfortunately, this error occurs quite often for me currently. Could be that #92 made things worse instead of better for this issue. Difficult to say with race conditions.

ce-bo:
Could be fixed by https://github.com/tree-sitter/node-tree-sitter/pull/75

ResistantBear:
It's already linked in the top description of this issue. However, taking a look on it again, that PR has been closed now without merge. 

https://github.com/tree-sitter/node-tree-sitter/pull/190#discussion_r1502694496

Looking at the PR that is linked there as it mentioned PR 75, the project seems to be in favour of blaming jest for this issue and linked https://github.com/jestjs/jest/issues/9206 for this. They seem to have agreed on not supporting jest at all for that reason. 

So we might solve this issue via #20 ... as that would mean to change all test cases an find an alternative or add additional libraries for replacing our current snapshot tests as such tests are not natively supported by node or mocha, it would be nice to verify that that would indeed solve the problem.

PR #110, opened 2024-02-13
Add special treatment for Java's init block #108

### LINKED ISSUE #108: JAVA Initialization block should be counted for complexity and functions metrics
Java initialization blocks contain logic, which would be executed when an instance is created or the class is loaded (static initialization block). They are similar to init block in Kotlin. Since we already count init block in Kotlin as complexity and functions, we should the same for Java initialization block.

mylinhdao:
Closes #108

### DIFF
--- a/src/parser/metrics/Complexity.ts
+++ b/src/parser/metrics/Complexity.ts
@@ -51,6 +51,7 @@ export class Complexity implements Metric {
     async calculate(parseFile: ParseFile, tree: Parser.Tree): Promise<MetricResult> {
         const queryBuilder = new QueryBuilder(parseFile, tree);
         if (parseFile.language === Languages.Java) {
+            //add query for instance init block in Java
             queryBuilder.setStatements(
                 this.mccStatementsSuperSet.concat(
                     new SimpleQueryStatement("(class_body (block)) @initBlock")
--- a/src/parser/metrics/Complexity.ts
+++ b/src/parser/metrics/Complexity.ts
@@ -8,8 +8,7 @@ import {
 } from "../helper/Model";
 import { FileMetric, Metric, MetricResult, ParseFile } from "./Metric";
 import { debuglog, DebugLoggerFunction } from "node:util";
-import { QueryMatch } from "tree-sitter";
-import Parser from "tree-sitter";
+import Parser, { QueryMatch } from "tree-sitter";
 import { Languages } from "../helper/Languages";
 
 let dlog: DebugLoggerFunction = debuglog("metric-gardener", (logger) => {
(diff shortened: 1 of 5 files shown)


src/parser/metrics/Complexity.ts
```diff
@@ -49,7 +51,17 @@ export class Complexity implements Metric {
...
+            case Languages.Java:
+                queryBuilder.setStatements(
+                    this.mccStatementsSuperSet.concat(
+                        new SimpleQueryStatement("(class_body (block)) @initBlock")
+                    )
+                );
+                break;
+            default:
+                queryBuilder.setStatements(this.mccStatementsSuperSet);
+        }
```
ResistantBear, line 64: @ce-bo suggested that we should not use switch-cases in a similar situation until we have more than one case to differentiate (see https://github.com/MaibornWolff/metric-gardener/pull/74#discussion_r1479348223), so we should also refactor this to an if-statement here.

src/parser/metrics/Functions.ts
```diff
@@ -19,7 +24,17 @@ export class Functions implements Metric {
...
+            case Languages.Java:
+                queryBuilder.setStatements(
+                    this.statementsSuperSet.concat(
+                        new SimpleQueryStatement("(class_body (block)) @initBlock")
+                    )
+                );
+                break;
+            default:
+                queryBuilder.setStatements(this.statementsSuperSet);
+        }
```
ResistantBear, line 37: The same applies here. Refactor to an if-statement.

ResistantBear (approved):
Seems like you are right that we actually need a special treatment here to find the non-static initialization blocks in Java. How unfortunate.

Looks fine, but please take care of the remarks.

src/parser/metrics/Functions.ts
mylinhdao: i changed to if else as suggested

src/parser/metrics/Complexity.ts
```diff
@@ -49,7 +50,16 @@ export class Complexity implements Metric {
...
+        if (parseFile.language === Languages.Java) {
+            //add query for instance init block in Java
+            queryBuilder.setStatements(
+                this.mccStatementsSuperSet.concat(
+                    new SimpleQueryStatement("(class_body (block)) @initBlock")
+                )
+            );
+        } else {
+            queryBuilder.setStatements(this.mccStatementsSuperSet);
+        }
```
ce-bo, line 62: Isn't it possible like this to shorten the code?

```suggestion
        if (parseFile.language === Languages.Java) {
            //add query for instance init block in Java
            this.mccStatementsSuperSet.concat(
                new SimpleQueryStatement("(class_body (block)) @initBlock")
            );
        }
        queryBuilder.setStatements(this.mccStatementsSuperSet);
```

ce-bo (approved):
In general LGTM :-) Thank you!

Unfortunately I was not aware that we need to add a special language handling for java init blocks. I would like to avoid that in general and to keep metric-gardener as simple and stupid as possible. Even if that means that some new or special language features are not supported (yet).

Let's merge this implementation because it does not add much complexity but let's discuss such cases for upcoming implementations.

src/parser/metrics/Complexity.ts
ResistantBear: No. `concat` actually returns a new array object that copies all elements. And that is on purpose, as mutating the class field (in this case `this.mccStatementsSuperSet`) like you seem to suggest here, e.g. by using `push` which mutates the existing array, would be a side effect that also affects further runs on other files. That was the exact issue that caused #65 and should be avoided therefore.

Issue #144, opened 2024-02-15
Performance Analysis


ResistantBear:
Analyse the performance of the file metric calculation of metric-gardener and may resolve "low hanging" performance issues before release.

ResistantBear:
I have created two flame graphs for running metric-gardener, on while parsing a Java and one while parsing a C++ code base.

Java:
![Image](https://github.com/MaibornWolff/metric-gardener/assets/34129499/f4153a33-cdbb-4c57-88bf-edc1791a79b3)

C++:
![Image](https://github.com/MaibornWolff/metric-gardener/assets/34129499/e6dfbb5c-9f90-4c54-88fe-0a8c7b927af8)

ResistantBear:
As can be seen, metric-gardener still spends most of the time in the `build()` method of `QueryBuilder` for both code bases. I have splitted out calling the constructor of `module:tree-sitter.Parser.Query` into the function `instantiateTreeSitterQuery` to pin down the problem to the actual constructor call. 

Another performance issue seems to be our new Real Lines of Code implementation with the recursive `walkTree()` function.

ResistantBear:
Flame graph for analyzing the CodeCharta code base:



![Image](https://github.com/MaibornWolff/metric-gardener/assets/34129499/7f46d44d-80b3-49d0-8a47-e9d1e9fc3dac)

ResistantBear:
Opened issue #148 for checking the possibility to cache tree sitter queries.

ResistantBear:
Opened issue #153 for looking into possible improvements for the real lines of code metric.

ResistantBear:
Opened separate issues #148 and #153 for possible mitigations of the two most relevant performance issues. So this is considered to be finished.

PR #154, opened 2024-02-16
Performance: Reuse queries to avoid calling the constructor

ResistantBear:
As shown in #144, the performance of the application is currently heavily impacted by calls to the constructor of `module:tree-sitter.Parser.Query`. Reusing these queries has therefore a huge potential to improve the performance. I implemented reusing queries via a cache functionality.

This reduces the time for analysing the CodeCharta-Repository from 17 seconds to 1,4 Seconds (-92%).

~~However, currently, this causes undeterministic failures where some Query.match(...) runs do return zero results.~~
According to [the documentation of the tree-sitter Query API](https://tree-sitter.github.io/tree-sitter/using-parsers#the-query-api), it should be fine to reuse queries.


mylinhdao (commented):
lets see if we can increase performance with these :) Im very excited !

src/parser/queries/QueryBuilder.ts
```diff
@@ -36,25 +42,57 @@ export class QueryBuilder {
...
+
+        if (cachedQueriesForLanguage === undefined) {
+            cachedQueriesForLanguage = new Map<string, CachedQuery>();
+            treeSitterQueryCache.set(this.#language, cachedQueriesForLanguage);
+        }
+        treeSitterQueryCache.get(this.#language)?.set(queryString, {
+            language: this.#language,
+            queryString: queryString,
+            query: newQuery,
+        });
```
ce-bo, line 83: I think you could change the order of the two blocks.
After adding a new query, you can be sure that there will be one for the current lookup. Pseudo Code:

```suggestion
        let cachedQueriesForLanguage = treeSitterQueryCache.get(this.#language);
        
        if (cachedQueriesForLanguage === undefined) {
            cachedQueriesForLanguage = new Map<string, CachedQuery>();
            treeSitterQueryCache.set(this.#language, cachedQueriesForLanguage);
        }
        
        const newQuery = new Query(languageToGrammar.get(this.#language), queryString);
        if (newQuery not in treeSitterQueryCache.get(this#language)) {
          treeSitterQueryCache.get(this.#language)?.set(queryString, {
              language: this.#language,
              queryString: queryString,
              query: newQuery,
          });
        }
        
      
            return cachedQueriesForLanguage.get(queryString);

```

ResistantBear: This change suggestion would call the constructor of Query every single time, which is the thing this change tries to avoid.

ResistantBear: But I could, instead of returning early, go into an `else`-block for creating and adding a new query and retrieve the query from the cache again at the end of the function. That would be an unessesary additional time of retrieving an element from the map, however, which I am hesitant to do on such a performance-critical part of the code.

mylinhdao, line 83: I think when we remove CachedQuery, the code could be more readable. What do you think?
```suggestion
        let queriesForLanguage = treeSitterQueryCache.get(this.#language);
        const query = queriesForLanguage?.get(queryString);

        if (queriesForLanguage !== undefined && query !== undefined)  return query;

        const newQuery = new Query(languageToGrammar.get(this.#language), queryString);

        if (queriesForLanguage === undefined) {
            treeSitterQueryCache.set(this.#language, new Map<string, Query>());
        }
        treeSitterQueryCache.get(this.#language)?.set(queryString, newQuery);
        return newQuery;
```

mylinhdao (commented):
I suggest some small changes to make the code more readable. Pls tell me your opinion :)

src/parser/queries/QueryBuilder.ts
```diff
@@ -7,14 +7,20 @@ let dlog: DebugLoggerFunction = debuglog("metric-gardener", (logger) => {
     dlog = logger;
 });
 
+const treeSitterQueryCache = new Map<Language, Map<string, CachedQuery>>();
+
+interface CachedQuery {
+    language: Language;
+    queryString: string;
+    query: Query;
+}
```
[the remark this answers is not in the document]
ResistantBear: Ah, yes, that was used for checking if there are hash collisions that cause the error I had here. That can be removed now.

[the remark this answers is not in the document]
ResistantBear: Done.

src/parser/queries/QueryBuilder.ts
ResistantBear: I have refactored that while removing the CachedQuery interface as suggested below.

ResistantBear: I have done something similar now

mylinhdao (approved):
LGTM now

Issue #148, opened 2024-02-16
May cache Tree Sitter Queries to avoid constructor call


ResistantBear:
The constructor call of the tree sitter query is very performance heavy. We might can avoid that by caching the queries. See also #144.

PR #167, opened 2024-02-19
Performance: reduce amount of recursive and API calls for real lines of code

### LINKED ISSUE #153: May improve the performance of the Real Lines Of Code (rloc) metric
As the flame graph from the performance analysis #144 shows, the calculation of this metric is quite performance-heavy. 
While most of that seems to be caused by calls to the tree-sitter API, we still might be able to improve on that.

ResistantBear:
This checks now if there is the necessity to traverse further down the tree before doing so. It also stores the result of `cursor.currentNode` instead of calling this tree-sitter API function multiple times.

When applied on top of the performance improvements of #148, this reduces the time to parse the CodeCharta codebase from 1,4 seconds to 0,72 seconds (720 milliseconds), approximately minus 50%

Resolves #153

### DIFF
--- a/src/parser/metrics/RealLinesOfCode.ts
+++ b/src/parser/metrics/RealLinesOfCode.ts
@@ -40,25 +40,33 @@ export class RealLinesOfCode implements Metric {
         isComment: (node: Parser.SyntaxNode) => boolean,
         realLinesOfCode = new Set<number>()
     ) {
+        const currentNode = cursor.currentNode;
         // This is not a comment syntax node, so assume it includes "real code".
-        if (!isComment(cursor.currentNode)) {
+        if (!isComment(currentNode)) {
             // Assume that first and last line of whatever kind of node this is, is a real code line.
             // This assumption should hold for all kinds of block/composed statements in (hopefully) all languages.
-            realLinesOfCode.add(cursor.startPosition.row);
+            realLinesOfCode.add(currentNode.startPosition.row);
 
             // This is a leaf node, so add further lines if this single token of actual code
             // spans over multiple lines and is no line ending:
-            if (cursor.currentNode.childCount == 0 && !"\r\n".includes(cursor.currentNode.type)) {
-                for (let i = cursor.startPosition.row + 1; i <= cursor.endPosition.row; i++) {
+            if (currentNode.childCount === 0 && !"\r\n".includes(currentNode.type)) {
+                for (
+                    let i = currentNode.startPosition.row + 1;
+                    i <= currentNode.endPosition.row;
+                    i++
+                ) {
                     realLinesOfCode.add(i);
                 }
             }
-
-            // Recurse, depth-first
-            if (cursor.gotoFirstChild()) {
-                this.walkTree(cursor, isComment, realLinesOfCode);
+            // Optimization: only go down the tree if we have not already covered all lines under this node:
+            else if (currentNode.endPosition.row > currentNode.startPosition.row) {
+                // Recurse, depth-first
+                if (cursor.gotoFirstChild()) {
+                    this.walkTree(cursor, isComment, realLinesOfCode);
+                }
             }
         }
+
         if (cursor.gotoNextSibling()) {
             this.walkTree(cursor, isComment, realLinesOfCode);
         } else {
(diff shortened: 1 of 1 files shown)


src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -40,25 +40,33 @@ export class RealLinesOfCode implements Metric {
         isComment: (node: Parser.SyntaxNode) => boolean,
         realLinesOfCode = new Set<number>()
     ) {
+        const currentNode = cursor.currentNode;
         // This is not a comment syntax node, so assume it includes "real code".
```
mylinhdao, line 44: we can delete this comment i think. It's clear enough what happens here.

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -40,25 +40,33 @@ export class RealLinesOfCode implements Metric {
...
-        if (!isComment(cursor.currentNode)) {
+        if (!isComment(currentNode)) {
             // Assume that first and last line of whatever kind of node this is, is a real code line.
```
mylinhdao, line 46: dont think we need this 2 comment lines

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -40,25 +40,33 @@ export class RealLinesOfCode implements Metric {
...
             // This assumption should hold for all kinds of block/composed statements in (hopefully) all languages.
-            realLinesOfCode.add(cursor.startPosition.row);
+            realLinesOfCode.add(currentNode.startPosition.row);
 
             // This is a leaf node, so add further lines if this single token of actual code
             // spans over multiple lines and is no line ending:
```
mylinhdao, line 51: How about:
```suggestion
            // Add more lines if a leaf node spans over multiple lines (except line-ending node)
```

mylinhdao (approved):
LGTM. But I think we should reduce comments where we can.

BridgeAR (approved):
RSLGTM

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -40,25 +40,33 @@ export class RealLinesOfCode implements Metric {
...
-
-            // Recurse, depth-first
-            if (cursor.gotoFirstChild()) {
-                this.walkTree(cursor, isComment, realLinesOfCode);
+            // Optimization: only go down the tree if we have not already covered all lines under this node:
+            else if (currentNode.endPosition.row > currentNode.startPosition.row) {
+                // Recurse, depth-first
+                if (cursor.gotoFirstChild()) {
+                    this.walkTree(cursor, isComment, realLinesOfCode);
+                }
```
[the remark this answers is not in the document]
ResistantBear: `gotoFirstChild()` returns false if there are no children, but it does not check whether the children span over additional code rows. There could be quite a lot of children like "(", "identifier", or ")" that are on the same row/do not span over following rows. We skip them now.

src/parser/metrics/RealLinesOfCode.ts
ResistantBear: I have removed most comments here now

src/parser/metrics/RealLinesOfCode.ts
```diff
@@ -40,25 +40,33 @@ export class RealLinesOfCode implements Metric {
         isComment: (node: Parser.SyntaxNode) => boolean,
         realLinesOfCode = new Set<number>()
     ) {
+        const currentNode = cursor.currentNode;
```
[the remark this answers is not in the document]
ResistantBear: I opened issue #172 for looking into that.

BridgeAR (commented):
Looking at this again: I am not certain if this actually works. If a node has children, it could also contain a comment?

mylinhdao (commented):
LGTM now

mylinhdao:
@BridgeAR  yes, a node with children can have a comment child node. But at the begin of the function `walkTree()`, we ensures that comment nodes are skipped with `if (!isComment(currentNode))`. 
The point for this PR is: when we find a node that covers only one line, then we ignore its children because they are on the sameline too. This reduces computing time. But if the node have multiple lines ( `currentNode.endPosition.row > currentNode.startPosition.row`), then we have to take a closer look at the children because it may contains comments lines :)

ResistantBear:
> Looking at this again: I am not certain if this actually works. If a node has children, it could also contain a comment?

Yes, but we do not have to care, as we already know that there is also actual code in the same line (the current node).

PR #171, opened 2024-02-19
Add support for Rust and Ruby programming languages

michaelalper-cezik-mw:
Imported Ruby & Rust tree-sitter grammars, added an extensive source file for both languages and adjusted nodeTypesConfig accordingly to assure basic support.

### DIFF
--- a/src/parser/config/nodeTypesConfig.json
+++ b/src/parser/config/nodeTypesConfig.json
@@ -7438,7 +7438,7 @@
     },
     {
         "expression": "singleton_method",
-        "metrics": [],
+        "metrics": ["functions"],
         "type": "statement",
         "category": "",
         "languages": ["rb"]
--- a/src/parser/config/nodeTypesConfig.json
+++ b/src/parser/config/nodeTypesConfig.json
@@ -7438,7 +7438,7 @@
     },
     {
         "expression": "singleton_method",
-        "metrics": ["functions"],
+        "metrics": ["functions", "complexity"],
         "type": "statement",
         "category": "",
         "languages": ["rb"]
--- a/src/parser/config/nodeTypesConfig.json
+++ b/src/parser/config/nodeTypesConfig.json
@@ -60,7 +60,7 @@
         "metrics": [],
         "type": "statement",
         "category": "",
-        "languages": ["cs", "go", "java", "cpp"]
+        "languages": ["cs", "go", "java", "cpp", "rs"]
     },
     {
         "expression": "indexer_declaration",
@@ -158,14 +158,14 @@
         "metrics": [],
         "type": "statement",
         "category": "",
-        "languages": ["cs", "java", "js", "php", "ts", "cpp", "tsx"]
+        "languages": ["cs", "java", "js", "php", "ts", "cpp", "tsx", "rs"]
     },
     {
         "expression": "await_expression",
         "metrics": [],
         "type": "statement",
         "category": "",
-        "languages": ["cs", "js", "ts", "tsx"]
+        "languages": ["cs", "js", "ts", "tsx", "rs"]
     },
     {
         "expression": "base_expression",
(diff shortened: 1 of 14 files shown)


mylinhdao (approved):
LGTM

src/parser/config/nodeTypesConfig.json
```diff
@@ -1948,10 +1948,10 @@
     },
     {
         "expression": "field_declaration_list",
-        "metrics": ["classes"],
```
ce-bo, line 1951: Was this a wrong mapping? If yes, would you please add a regression test? At least one test should have failed in my opinion.

src/parser/config/nodeTypesConfig.json
```diff
@@ -7036,5 +7036,1433 @@
...
+    {
+        "expression": "singleton_class",
+        "metrics": [],
+        "type": "statement",
+        "category": "",
+        "languages": ["rb"]
+    },
+    {
+        "expression": "singleton_method",
+        "metrics": [],
```
ce-bo, line 7441: Are these also relevant for functions metric?

src/parser/config/nodeTypesConfig.json
ResistantBear: If I remember correctly, the issue was that this did work for C++, but not for Rust. As `class_specifier` and `struct_specifier` should be the more straight forward choice for C++ anyway, we decided to count them instead for C++.

ResistantBear: I do not know about Go, as there were no test cases provided by you for the classes metric for that language while mentioning the reason that there are no classes in Go. But I see now that we actually counted this as class for Go, mh. We might look into that again.

test/parser/RustMetrics.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    describe("parses Rust real lines of code metric", () => {
+        it("should count correctly for a non-empty file, ignoring comments and empty lines", () => {
+            testFileMetric(
+                rsTestResourcesPath + "rust-example-code.rs",
+                FileMetric.realLinesOfCode,
+                70
+            );
+        });
+    });
+});
```
ResistantBear, line 61: There is no tests for the functions metric.

test/parser/RubyMetrics.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    describe("parses Ruby real lines of code metric", () => {
+        it("should count correctly for a non-empty file, ignoring comments and empty lines", () => {
+            testFileMetric(
+                rbTestResourcesPath + "ruby-example-code.rb",
+                FileMetric.realLinesOfCode,
+                62
+            );
+        });
+    });
+});
```
ResistantBear, line 57: There is no tests for the functions metric.

src/parser/config/nodeTypesConfig.json
ResistantBear: I have checked. We should indeed count them. Accordingly, I have added a mapping to functions and complexity as well as test cases.

src/parser/config/nodeTypesConfig.json
```diff
@@ -7036,5 +7036,1433 @@
...
+    {
+        "expression": "retry",
+        "metrics": [],
+        "type": "statement",
+        "category": "",
+        "languages": ["rb"]
+    },
+    {
+        "expression": "singleton_class",
+        "metrics": [],
```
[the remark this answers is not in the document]
ResistantBear: It seems like this gets generated if we have a `class << self` construct in Ruby. I have only seen examples for this inside of a class definition for defining something like static methods to the surrounding class. In this case, not counting this token yields the expected result, as adding static methods to a class is no new class definition.

test/parser/RubyMetrics.test.ts
ResistantBear: I have added that myself now.

ResistantBear (changes requested):
Adding the max_nesting_level metric is blocked until #179 is discussed and implemented.

ResistantBear (approved):

PR #174, opened 2024-02-19
Enforce reading the file before using the language of a file

### LINKED ISSUE #166: Bug: restructure file reading/tree parsing to be guaranteed to happen before any other operation requiring the Language
In the TreeParser.ts, we may change the grammar used for the file to TSX based on the file content. We however still use queries based upon the Information in the ParseFile, which is still only based upon the file extension. We have to make sure that

1. We execute TreeParser/the reading of the file before any query is build for that file.
2. We set the language in the ParseFile for that file.

ResistantBear:
Does enforce that the source code file is read before any code can make use of the language field of `ParseFile` (now renamed to `ParsedFile` to reflect that the tree for this has been already parsed). This refactors the code accordingly while keeping the reading of files and parsing of file metrics asynchonous.

This change avoids that the language "JavaScript" assumed via the file extension is just used for creating queries, despite that the tree on which these queries are run could be parsed with the TSX grammar after the actual source code is evaluated. This change should also allow to handle similar situations with other languages if they arise.

Closes #166


src/parser/GenericParser.ts
```diff
@@ -38,40 +45,67 @@ export class GenericParser {
...
+            for await (const file of filePathGenerator) {
+                const fileExtensionLanguage = fileExtensionToLanguage.get(file.fileExtension);
+                if (fileExtensionLanguage !== undefined) {
+                    parsePromises.set(
+                        file.filePath,
+                        TreeParser.parse(file, fileExtensionLanguage).catch((reason) => {
+                            console.error(
+                                "Error while parsing a tree for the file " + file.filePath
+                            );
+                            console.error(reason);
```
ce-bo, line 60: This is indead something we should display to the users, so that they are aware that some files could not be parsed.

src/parser/GenericParser.ts
```diff
@@ -38,40 +45,67 @@ export class GenericParser {
...
-                        unknownParseFiles.push(file);
-                    }
+                for (const [filePath, parsePromise] of parsePromises) {
+                    fileMetricPromises.push(
+                        metricsParser.calculateMetrics(parsePromise).catch((reason) => {
+                            console.error("Error while parsing file metrics");
+                            console.error(reason);
+                            return [
+                                filePath,
+                                new Map([["ERROR", { metricName: "ERROR", metricValue: -1 }]]),
```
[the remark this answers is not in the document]
ce-bo: That is a good idea. we should think about on how to display errors in general. First I think it is good that a user can see errors in the terminal. It does make sense to also include them in an error/info attriubte in the output file.

src/parser/MetricCalculator.ts
```diff
@@ -45,42 +44,45 @@ export class MetricCalculator {
...
+    async calculateMetrics(
+        parsedFilePromise: Promise<SimpleFile>
+    ): Promise<[string, Map<string, MetricResult>]> {
+        const parsedFile = await parsedFilePromise;
 
-            dlog(" ------------ Parsing File " + parseFile.filePath + ":  ------------ ");
+        if (!isParsedFile(parsedFile)) {
+            throw new Error(
+                "Unable to calculate file metrics because there was an error while creating the tree."
+            );
```
ce-bo, line 59: Adapt error handling (see comments above).

src/parser/GenericParser.ts
```diff
@@ -38,40 +45,67 @@ export class GenericParser {
         let couplingMetrics = {} as CouplingResult;
 
         try {
-            const parseFilesGenerator = findFilesAsync(this.config);
+            const filePathGenerator = findFilesAsync(this.config);
```
ce-bo, line 48: I think we should add unit tests to be sure that this code behaves as expected on success, on errors, edgecaes etc.

src/parser/MetricCalculator.ts
ResistantBear: This error is catched by the error handling on which you commented above. So I expect it is fine?

src/parser/GenericParser.ts
ResistantBear: For the findFilesAsync-function? I can do that.
For GenericParser? Well, all our current test cases originated from your unit tests for `GenericParser.calculate`, that function is basically our main function that initializes and starts everything related to calculating metrics. I could add some more test cases with files that might trigger errors?

ResistantBear: Implemented Unit-tests for `GenericParser.calculate` by mocking all more complex called functions within it. 
See `src/parser/GenericParser.test.ts`.

src/parser/GenericParser.ts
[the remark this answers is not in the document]
ResistantBear: I guess this is out of scope of this PR and should be discussed together with #179.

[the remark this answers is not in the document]
ResistantBear: As we already discussed to keep the "info" field, I will implement this.

src/parser/GenericParser.test.ts
```diff
@@ -0,0 +1,281 @@
...
+
+describe("GenericParser.calculateMetrics() unit tests", () => {
+    it("should return the correct value and call the appropriate functions in case there is no error", async () => {
+        const expectedFileMetricsMap = new Map([
+            [FileMetric.linesOfCode, { metricName: FileMetric.linesOfCode, metricValue: 5 }],
+        ]);
+
+        const findFilesAsyncMock = findFilesAsync as jest.Mocked<typeof findFilesAsync>;
+        findFilesAsyncMock.mockImplementation((config) => {
+            return mockedFindFilesAsync();
```
ce-bo, line 53: Seems to be something like `findFilesWIthInvalidPathAsync`

src/parser/GenericParser.test.ts
```diff
@@ -0,0 +1,281 @@
...
+    tree = parser.parse("int main() { return 0; }");
+});
+
+beforeEach(() => {
+    // Clear all instances and calls to constructor and all methods:
+    jest.clearAllMocks();
+    jest.resetAllMocks();
+});
+
+describe("GenericParser.calculateMetrics() unit tests", () => {
```
ce-bo, line 45: The combination of describe and it blocks can help you to have a human readable message in the test outputs. In this case it would be `GenericParser.calculateMetrics() unit tests should return the correct value and call the appropriate functions in case there is no error`. Maybe it would be better to have something like describe: `GenericParser`, it: `should calculate lines of code metric`

src/parser/GenericParser.test.ts
```diff
@@ -0,0 +1,281 @@
...
+    it("should return the correct value and call the appropriate functions in case there is no error", async () => {
```
ce-bo, line 46: It is hard to know from this title what this unit test is actually doing. Let's describe via the should when pattern what happens. Example: `should calculate lines_of_code properly when valid file is parsed`

src/parser/GenericParser.test.ts
```diff
@@ -0,0 +1,281 @@
...
+        const actualResult = await parser.calculateMetrics();
+
+        expect(actualResult.fileMetrics).toEqual(
+            new Map([["clearly/invalid/path.cpp", expectedFileMetricsMap]])
+        );
+        expect(calculateMetricsSpied).toHaveBeenCalledTimes(1);
+        expect(treeParserSpied).toHaveBeenCalledTimes(1);
+    });
+
+    it("should return the appropriate values and call the appropriate functions in case there is no error and two files are found", async () => {
```
ce-bo, line 81:
```suggestion
    it("should calculate metric properly when multiple valid files are parsed", async () => {
```

src/parser/GenericParser.test.ts
```diff
@@ -0,0 +1,281 @@
...
+            /* Do nothing */
+        });
+
+        const parser = new GenericParser(getParserConfiguration("clearly/invalid/path.cpp"));
+        const actualResult = await parser.calculateMetrics();
+
+        // TODO: do no longer include errors into the list of results, use "info" field instead.
+        expect(actualResult.fileMetrics).toEqual(
+            new Map([["clearly/invalid/path.cpp", expectedFileMetricsMap]])
+        );
```
ce-bo, line 151: I like to have // given // when // then comment within the tests to have a clearer structure. 

```suggestion
        // given
        const findFilesAsyncMock = findFilesAsync as jest.Mocked<typeof findFilesAsync>;
        findFilesAsyncMock.mockImplementation((config) => {
            return mockedFindFilesAsync();
        });

        jest.spyOn(TreeParser, "parse").mockImplementation(async (file, assumedLanguage) => {
            throw new Error("Baaaaaah");
        });

        jest.spyOn(MetricCalculator.prototype, "calculateMetrics").mockResolvedValue([
            "clearly/invalid/path.cpp",
            expectedFileMetricsMap,
        ]);

        const errorSpy = jest.spyOn(console, "error").mockImplementation(() => {
            /* Do nothing */
        });

        const parser = new GenericParser(getParserConfiguration("clearly/invalid/path.cpp"));
        
        // when
        const actualResult = await parser.calculateMetrics();

        // then
       const expectedFileMetricsMap = new Map([
            ["ERROR", { metricName: "ERROR", metricValue: -1 }],
        ]);
        expect(actualResult.fileMetrics).toEqual(
            new Map([["clearly/invalid/path.cpp", expectedFileMetricsMap]])
        );
```

The given block contains your setup and mocks
The when block contains the function or action that is executed
Then contains assertions.

src/parser/GenericParser.test.ts
```diff
@@ -0,0 +1,281 @@
...
+        expect(errorSpy).toHaveBeenCalled();
+    });
+
+    it("should return the appropriate value and call the appropriate functions in case there is an error while calculating metrics", async () => {
```
ce-bo, line 155: Let's try to be more explicit: What is an appropriate value? What it the appropariate action/function that is called? Which type of error is thrown?

src/parser/GenericParser.test.ts
```diff
@@ -0,0 +1,281 @@
...
+        expect(actualResult.fileMetrics).toEqual(
+            new Map([
+                ["clearly/invalid/path1.cc", expectedErrorMetricsMap],
+                ["clearly/invalid/path2.cpp", expectedFileMetricsMap],
+            ])
+        );
+        expect(errorSpy).toHaveBeenCalled();
+    });
+
+    it("should fail gracefully if findFilesAsync throws an error", async () => {
```
ce-bo, line 247: A general note: If there is a lot of copied test code: Tests should be handled like production code. Let's create helper methods if it is helpful, let's move reusable code into beforeEach methods, etc.

src/parser/GenericParser.test.ts
```diff
@@ -0,0 +1,281 @@
...
+        });
+
+        jest.spyOn(MetricCalculator.prototype, "calculateMetrics").mockResolvedValue([
+            "clearly/invalid/path.cpp",
+            expectedFileMetricsMap,
+        ]);
+
+        const errorSpy = jest.spyOn(console, "error").mockImplementation(() => {
+            /* Do nothing */
+        });
```
ce-bo, line 143:
```suggestion
        mockConsoleErrorDoNothing();
```

src/parser/GenericParser.test.ts
```diff
@@ -0,0 +1,281 @@
...
+
+        jest.spyOn(TreeParser, "parse").mockImplementation(async (file, assumedLanguage) => {
+            throw new Error("Baaaaaah");
+        });
+
+        jest.spyOn(MetricCalculator.prototype, "calculateMetrics").mockResolvedValue([
+            "clearly/invalid/path.cpp",
+            expectedFileMetricsMap,
+        ]);
```
ce-bo, line 139:
```suggestion
        mockCalculateMetrics(path, metricsToBeReturned);
```

src/parser/GenericParser.test.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    it("should return the appropriate value and call the appropriate functions in case there is an error while parsing the tree", async () => {
+        const expectedFileMetricsMap = new Map([
+            ["ERROR", { metricName: "ERROR", metricValue: -1 }],
+        ]);
+
+        const findFilesAsyncMock = findFilesAsync as jest.Mocked<typeof findFilesAsync>;
+        findFilesAsyncMock.mockImplementation((config) => {
+            return mockedFindFilesAsync();
+        });
```
ce-bo, line 130: Could be done in a beforeEach I guess
```suggestion
        mockFindFilesAsync()
```

src/parser/GenericParser.test.ts
ResistantBear: Mh, as it does not actually find something but just fakes it, I find "mocked...." more clear.

src/parser/GenericParser.test.ts
ResistantBear: Changed the naming

src/parser/GenericParser.test.ts
ResistantBear: We mock `metricsParser.calculateMetrics(...)` here, so we do not actually calculate anything nor check the validity of the file. We just test if `GenericParser.calculateMetrics` performs the actions it should

michaelalper-cezik-mw (approved):

PR #180, opened 2024-02-22
fix: classes metric for go

### LINKED ISSUE #175: Count structs and interfaces in Go
Currently, we do not test the classes metric for Go. However, there are Interfaces and Structs in the Go programming language that we should count for that metric. 

Task:

- [x] Provide Go test source file(s) that include structs and interfaces
- [x] Write test cases for the classes metric for Go
- [x] Adjust nodeTypesConfig.json to count structs and interfaces of Go for the classes metric

ResistantBear:
The classes metric is now tested and calculated correctly for the Go programming language.
Does no longer count "field_declaration_list" as class for Go and C++, as that lead to duplications.

Fixes #175


michaelalper-cezik-mw (approved):
Looks good

PR #192, opened 2024-02-29
Support for the C programming language, support for C/C++ Enums and Unions, various Bugfixes

### LINKED ISSUE #189: Add support for C
Add support for the C programming language using tree-sitter-c

NOTE: As .h files might be used for C or C++, we should provide an option to explicitely state whether we should parse files as C or C++. Per default, we should parse .h files as C++, as C++ is (mostly) a superset of C. 
We should also consider that .C files (capital C) should default to C++ but .c files should default to C, just like gcc does. Currently, or file extension handling is not case sensitive.

ResistantBear:
Adds support for the C programming language

Includes the following further changes:

- New command line options `parse-h-as-c` and `parse-some-h-as-c` to decide that (some) .h header files should be parsed as C files instead of C++ files. C++ remains the default for .h files, as it is mostly a superset of C.
- Refactored the Configuration constructor to make parameters more explicit
- Added documentation for all available command line options to the Readme, removed outdated information.
- Added system-independent unit tests for the new `assumeLanguageFromFilePath(...)` function that realizes the new command line options.
- Added support and test cases for C and C++ enums and unions.
- Added special queries for the C and C++ classes metric to avoid the duplicated counting of classes, structs, enums or unions when using `typedef` statements like `typedef class name alias` or forward declarations like `class A; class A {};`. Also considers opaque enums.
- Added test cases for the typedef and forward declaration issues mentioned above.
- Fixes a bug in the `formatPrintPath(...)` function of Helper.ts that resulted in an empty file path when the user supplied sources path is a path to a single file.
- Added system-independent unit tests for `formatPrintPath(...)` to avoid regression, tested for both UNIX (/.../.../...) and DOS-style (C:\\...\\...\\...) paths.

Closes #189

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,9 +54,7 @@ This is quite slow and can take up to one or two hours but can provide good resu
 -   Ruby
 -   Rust
 -   Bash (the binary logical operators -o for "or" and -a for "and" are currently not evaluated due to issues with tree-sitter-bash. For the same reason, the && and || operators are not evaluated if placed after the first heredoc delimiter. Default labels in Switch-statement are treated as regular case labels.)
--   C
-
-Note: to parse .h files as C instead of C++, use the command line option `parse-h-as-c` or `parse-some-h-as-c`.
+-   C (take a look at the CLI option [`parse-h-as-c`](#command-line-options-for-the-parse-command) or [`parse-some-h-as-c`](#command-line-options-for-the-parse-command) for parsing `.h` C headers as C)
 
 ### Supported File Metrics
 
@@ -90,19 +88,32 @@ The number of comment lines inside a file. Does count for any kind of comment (e
 **real_lines_of_code**<br>
 The number of lines inside a file that contain actual code of the programming language, not counting for comments, empty lines, etc.
 
-### Coupling Metrics (experimental)
+### Command line options for the `parse` command
 
-**Supported for C# and PHP**<br>
-Activate dependency analysis by passing `--parseDependencies`
+`--help`<br>
+Shows an overview over all options and their usage
 
--   Coupling Between Objects (CBO)
--   Incoming Dependencies
--   Outgoing Dependencies
--   Instability: Outgoing Dependencies / (Outgoing Dependencies + Incoming Dependencies)
+`--output-path`, `-o`<br>
+Specifies the location on which the .json-file with the results should be stored, e.g. `./metrics.json`. This is required.
 
-**Limitations:**<br>
+`--relative-paths`, `-r`<br>
+Write paths to the analyzed files in the output .json-file relative to the source directory instead of writing the absolute paths to the files. E.g. writes `src/package1/file1.java` instead of `/home/user.name/programming/project1/src/package1/file1.java` (works also for Windows-style paths).
 
--   Multiple, nested Namespace Declarations within one .cs file are not covered so far and are ignored during the calculation of coupling.
+`--exclusions`, `-e`<br>
+Excludes the specified folders from being scanned for files to be analyzed. Can be an arbitrarily nested subdirectory, so something like `--exclusions "folder1"` does exclude `src/package1/folder1`. Does exclude all files in the folder itself as well as in all subdirectories of the folder.
+
+`--parse-h-as-c`, `--hc`<br>
+Parse all files with the file extension `.h` as files written in the C programming language instead of C++. Defaults to C++, as the C++ grammar is mostly a superset of the C grammar.
+
+`--parse-some-h-as-c`, `--shc`<br>
+Parse all files inside the specified folders that have the file extension `.h` as files written in the C programming language instead of C++. Defaults to C++. Can be also used for file names, e.g. `--shc header-file.h`. Similar to `--exclusions`, the folder/file can lie in an arbitrary subdirectory and the option applies to all files in all subdirectories of the specified folder(s).
+
+`--compress`, `-c`<br>
+Compresses the output .json-file into an zip-archive.
+
+`--parse-dependencies`<br>
+EXPERIMENTAL - VERY SLOW AND NOT READY FOR PRODUCTIVE USE YET<br>
+Performs a dependency analysis and appends the results to the output .json-file (see [below](#experimental-coupling-metrics)).
 
 ### Updating tree-sitter grammars and adding support for more languages
 
@@ -117,6 +128,22 @@ Check out our contribution guidelines in the file [CONTRIBUTING.md](CONTRIBUTING
 There are additional outputs about the metric calculation process that can be enabled by setting the
 environment variable `NODE_DEBUG` to `metric-gardener`.
 
+### Experimental Coupling Metrics
+
+**EXPERIMENTAL - VERY SLOW AND NOT READY FOR PRODUCTIVE USE YET**
+
+**Supported for C# and PHP**<br>
+Activate dependency analysis by passing `--parseDependencies`
+
+-   Coupling Between Objects (CBO)
+-   Incoming Dependencies
+-   Outgoing Dependencies
+-   Instability: Outgoing Dependencies / (Outgoing Dependencies + Incoming Dependencies)
+
+**Limitations:**<br>
+
+-   Multiple, nested Namespace Declarations within one .cs file are not covered so far and are ignored during the calculation of coupling.
+
 ### TODOs
 
 Next Steps:
@@ -127,7 +154,6 @@ Next Steps:
 
 Performance Optimizations:
 
--   Mapped Expressions: Query expressions for current language only instead of brute force all available metric expressions
 -   Skip primitive types like (void, boolean, etc.) during accessor scan and usage candidates building
 -   Exclude System Namespaces like System in CSharp etc. (Configurable option would also be nice (e.g. exclude NameSpace _UnitTest_))
 -   Improve performance in Abstract Usage Collector (add candidates only if they can be found in previously retrieved fully qualified type names in other words the Namespace Collector).
@@ -145,7 +171,6 @@ Other TODOs and ideas:
 -   Performance & Duplicate Adds (see TODO comments)
 -   Export relationships as a graph file
 -   Write command to add new language and map expressions
--   Checkout sample project(s) per language and parse them as an automatically test
+-   Checkout sample project(s) per language and parse them as an automatic test
 -   ProgressBar
 -   Separate Infrastructure from Domain Code
--   Support more languages
--- a/README.md
+++ b/README.md
@@ -62,7 +62,6 @@ Counts expressions that branch the control flow, like if-statements, loops, swit
 The number of function declarations inside a file. Includes all kinds of functions, like constructors, lambda functions, member functions, abstract functions, constructors, init blocks, closures, etc.
 
 **classes**<br>
-<<<<<<< HEAD
 The number of class definitions inside a file, also counting for enums, interfaces, structs, unions, traits and records.
 
 <details><summary>Language specific details</summary>
--- a/README.md
+++ b/README.md
@@ -9,12 +9,6 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 **It is based on grammars from [tree-sitter](https://github.com/tree-sitter/tree-sitter).**
 
-### Main Issues with MetricGardener
-
-1.  It is not very well tested
-2.  Because of 1., there might be some issues for certain languages
-3.  The list of supported languages is still rather limited
-
 ### Usage
 
 #### Install required build tools:
--- a/README.md
+++ b/README.md
@@ -67,7 +67,8 @@ Counts expressions that branch the control flow, like if-statements, loops, swit
 The number of function declarations inside a file. Includes all kinds of functions, like constructors, lambda functions, member functions, abstract functions, constructors, init blocks, closures, etc.
 
 **classes**<br>
-The number of class definitions inside a file, also counting for enums, interfaces, structs, traits, records and types.
+<<<<<<< HEAD
+The number of class definitions inside a file, also counting for enums, interfaces, structs, unions, traits and records.
 
 <details><summary>Language specific details</summary>
 
--- a/README.md
+++ b/README.md
@@ -54,6 +54,9 @@ This is quite slow and can take up to one or two hours but can provide good resu
 -   Ruby
 -   Rust
 -   Bash (the binary logical operators -o for "or" and -a for "and" are currently not evaluated due to issues with tree-sitter-bash. For the same reason, the && and || operators are not evaluated if placed after the first heredoc delimiter. Default labels in Switch-statement are treated as regular case labels.)
+-   C
+
+Note: to parse .h files as C instead of C++, use the command line option `parse-h-as-c` or `parse-some-h-as-c`.
 
 ### Supported File Metrics
 
--- a/resources/c++/enums_and_unions.cpp
+++ b/resources/c++/enums_and_unions.cpp
@@ -1,21 +1,21 @@
 /*
  * Unscoped enums:
  */
-enum unscoped_enum { Kartoffel, Karotte, Kohlrabi, Krautsalat };
-enum unscoped_fixed_type_enum: int { Eins, Zwei, Drei };
+enum unscoped_enum { Potato, Carrot, Kohlrabi, Coleslaw };
+enum unscoped_fixed_type_enum: int { One, Two, Tree };
 // unscoped enums with unfixed type cannot be opaque/forward declared.
 enum unscoped_fixed_type_opaque_enum: int;
 
 /*
  * Scoped enums:
  */
-enum class scoped_class_enum { Kartoffel, Karotte, Kohlrabi, Krautsalat };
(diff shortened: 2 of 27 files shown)


README.md
```diff
@@ -54,6 +54,9 @@ This is quite slow and can take up to one or two hours but can provide good resu
 -   Ruby
 -   Rust
 -   Bash (the binary logical operators -o for "or" and -a for "and" are currently not evaluated due to issues with tree-sitter-bash. For the same reason, the && and || operators are not evaluated if placed after the first heredoc delimiter. Default labels in Switch-statement are treated as regular case labels.)
+-   C
+
+Note: to parse .h files as C instead of C++, use the command line option `parse-h-as-c` or `parse-some-h-as-c`.
```
ce-bo, line 59: It could be enough to specify only one parameter parse-h-as-c. A list of files can be provided optionally. If there is no list of file paths all header files are counted as c.

src/parser/helper/Language.ts
```diff
@@ -102,6 +109,57 @@ export const fileExtensionToLanguage = new Map([
...
+export function assumeLanguageFromFilePath(
+    filePath: string,
+    config: Configuration,
+    pathModule = path,
+) {
+    const fileExtension: string = getFileExtension(filePath);
+
+    // Handling of the parse .h as C option:
+    if (fileExtension === "h") {
+        if (config.parseAllAsC) {
```
ce-bo, line 137:
```suggestion
        if (config.parseAllHeaderFilesAsC) {
```

resources/c++/enums_and_unions.cpp
```diff
@@ -0,0 +1,69 @@
+/*
+ * Unscoped enums:
+ */
+enum unscoped_enum { Kartoffel, Karotte, Kohlrabi, Krautsalat };
```
ce-bo, line 4: I prefer English examples please :-)

resources/c++/typedefs.h
```diff
@@ -0,0 +1,83 @@
...
+enum Food { Kartoffel, Sauerkraut, Bratwurst, Brezel, Bauernbrot };
+
+// Scoped enum:
+enum struct Scoped_Enum { Pizza, Pasta, Schnitzel, Backfisch };
+// Opaque scoped enum:
+enum struct Scoped_Enum_2: int;
+
+// Should not count as enum, as it defines just an alias.
+typedef enum Food Food;
+// The following does not work:
```
ce-bo, line 13: What do you mean by that exactly?

src/app.ts
```diff
@@ -29,30 +29,44 @@ yargs(hideBin(process.argv))
...
-                .option("compress", {
-                    alias: "c",
+                .option("parse-h-as-c", {
+                    alias: "hc",
                     type: "boolean",
-                    description: "output .gz-zipped file",
+                    description: "Parse all .h files as C instead of C++ (defaults to C++)",
                     default: false,
                 })
+                .option("parse-some-h-as-c", {
```
ce-bo, line 44: We should provide a usage example in the readme and explain if it the option applies recursively for the specified paths.

src/parser/Configuration.ts
```diff
@@ -69,7 +82,23 @@ export class Configuration {
         this.outputPath = outputPath;
         this.parseMetrics = true;
         this.parseDependencies = parseDependencies;
-        this.exclusions = new Set(exclusions.split(",").map((exclusion) => exclusion.trim()));
+
+        if (exclusions.length > 0) {
+            this.exclusions = new Set(exclusions.split(",").map((exclusion) => exclusion.trim()));
+        } else {
+            this.exclusions = new Set();
+        }
```
ce-bo, line 90: You could use a ternary here.

src/parser/helper/Helper.test.ts
```diff
@@ -0,0 +1,140 @@
...
+                "/some/path",
+                "invalid/output/path",
+                false,
+                "",
+                false,
+                "",
+                false,
+                true,
+                true,
+            );
```
ce-bo, line 133: It is hard to see for what which argument is.
Maybe this could help::
const config = `{ sourcePath: "a/b/c", otherAtrribute: false } as Configuration`. 

I am not sure though if this compiles :-)

src/parser/helper/Helper.test.ts
```diff
@@ -0,0 +1,140 @@
...
+                "",
+                false,
+                true,
+                false,
+            );
+
+            expect(formatPrintPath(filePath, config, path.win32)).toEqual("file.extension");
+        });
+
+        it("should both create relative paths and replace forward slashes when configured", () => {
```
ce-bo, line 121: Let's have a more descriptive `when` in the test name
```suggestion
        it("should both create relative paths and replace forward slashes when enforceBackwardSlashes is set", () => {
```

test/metric-end-results/TestHelper.ts
```diff
@@ -16,8 +17,9 @@ import { strcmp } from "../../src/parser/helper/Helper";
  * results.fileMetrics.get(formatPrintPath(inputPath, config))
  * </code></pre>
  */
-export function getParserConfiguration(
+export function getParserTestConfiguration(
```
ce-bo, line 20: There is a way to replace single attributes of a default configuration object like this:
var config = {...defaultConfig, ...customConfig}; 

Attributes with the same name are overwritten by the ones of the customConfig. By that you can pass something like this:
```
const customConfig = { enforceBackwardSlashes:: false}
getParserTestConfiguration(customConfig: Partial<Configuration>): Configuration {
const defaultConfig = // .. implementation;

return {...defaultConfig, ...customConfig};
```

src/parser/metrics/Metric.ts
```diff
@@ -79,28 +79,22 @@ export abstract class SourceFile {
...
-    constructor(filePath: string, fileExtension: string) {
-        super(filePath, fileExtension);
+    constructor(filePath: string) {
+        super(filePath);
     }
 }
 
 /**
- * Represents a supported file to be analyzed for metrics, including its path, file extension, language and parsed syntax tree.
+ * Represents a supported file to be analyzed for metrics, including its path, language and parsed syntax tree.
```
ce-bo, line 97: Do you think it is really necessary to mention all the attributes in the comment? This is a coupling and you have to update the comment all the time.

src/parser/helper/Language.ts
```diff
@@ -102,6 +109,57 @@ export const fileExtensionToLanguage = new Map([
...
+ */
+export function assumeLanguageFromFilePath(
+    filePath: string,
+    config: Configuration,
+    pathModule = path,
+) {
+    const fileExtension: string = getFileExtension(filePath);
+
+    // Handling of the parse .h as C option:
+    if (fileExtension === "h") {
```
ce-bo, line 136: Let's have an early return here: if (!isHeaderFile(fileExtension)) { return; }

src/parser/helper/Language.ts
```diff
@@ -102,6 +109,57 @@ export const fileExtensionToLanguage = new Map([
...
+        }
+        return Language.CPlusPlus;
+    }
+
+    let result = caseSensitiveFileExtensionToLanguage.get(fileExtension);
+    const inLowerCase = fileExtension.toLowerCase();
+    if (result === undefined) {
+        result = fileExtensionToLanguage.get(inLowerCase);
+    }
+    return result;
```
ce-bo, line 160: I think this lookup logic will most likely be reused in other parts. To extract this into a helper function would make sense.

src/parser/metrics/Classes.ts
```diff
@@ -15,6 +18,58 @@ export class Classes implements Metric {
 
     constructor(allNodeTypes: ExpressionMetricMapping[]) {
         this.statementsSuperSet = getQueryStatements(allNodeTypes, this.getName());
+
+        this.addQueriesForC_Cpp();
+    }
+
+    addQueriesForC_Cpp() {
```
ce-bo, line 25:
```suggestion
    addQueriesForCAndCpp() {
```

resources/c++/enums_and_unions.cpp
ResistantBear: I thought that would be more funny compared to "entry1, entry2, entry3" or something. Can't we have some nice things like Kartoffeln? :D

resources/c++/typedefs.h
ResistantBear: It does not compile. I will make that more specific.

src/app.ts
```diff
@@ -29,30 +29,44 @@ yargs(hideBin(process.argv))
...
+                    alias: "c",
+                    type: "boolean",
+                    description: "output .gz-zipped file",
+                    default: false,
+                })
+                .option("parse-dependencies", {
+                    type: "boolean",
+                    default: false,
+                    description:
+                        "Flag to enable dependency parsing (dependencies will be appended to the output file)",
```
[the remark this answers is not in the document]
ResistantBear: I can do that. I have only changed to order of options here, so that parse-dependencies is less visible.

src/parser/helper/Language.ts
ResistantBear: Well, not all, but only the once with the file extension `.h`. Thats not the only file extension for headers, there are also `.hpp`, `.hh`, `.hxx` or `.H` (probably even more).
I struggled there because I wanted to name it `parseAllHAsC` just like the command line option "parse-h-as-c", but that does really not work and is completely not comprehensible in camel case.

src/app.ts
ResistantBear: It does, just like the "exclusions" option. Speaking of which, we should probably also include an usage example, etc., for that option?

src/parser/Configuration.ts
ResistantBear: I could. I prefer to not write "one line that does it all". But I can change that here.

src/parser/helper/Helper.test.ts
ResistantBear: I will look into it.

src/parser/metrics/Metric.ts
ResistantBear: No, I can change that.

src/parser/helper/Language.ts
```diff
@@ -102,6 +109,57 @@ export const fileExtensionToLanguage = new Map([
...
+            // that higher-level folders are evaluated for this:
+            const relativePath = pathModule.relative(config.sourcesPath, filePath);
+            const backwardSlashRelpath = replaceForwardWithBackwardSlashes(relativePath);
+            const relpathSplitted = backwardSlashRelpath.split("\\");
+            for (const pathElement of relpathSplitted) {
+                if (config.parseSomeAsC.has(pathElement)) {
+                    return Language.C;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ResistantBear: It is only done for .h files and only if the new option is used, so I guessed that it is neglectable. 
The only other option would be to do that when the files are found, but that would mean refactoring that part again, as I removed checking the language while finding the file to handle cases like the @flow annotation that can change the language of the file and can be only registered after reading the file content. Having the language "found out" at one place and changed again at another place while ensuring that the language of a file has not been used before reading the file would be very error prone.

test/metric-end-results/TestHelper.ts
ResistantBear: Interesting. I will look into that.

README.md
ResistantBear: That would be nice, but it does not seem to work with the yargs library used here. If I change "parse-h-as-c" to type "string" just as the "parse-some-h-as-c" parameter, it does unfortunately still provide the specified default value when using the option while not providing any purther input. So `parse sources -o output.json` is, for the program, indistinguishable from `parse sources -o output.json --parse-h-as-c`

src/app.ts
ce-bo: Yes, but I think the exclusion feature is not really nice right now. We should rework this to accept exclusion patterns. But this is out of scope here.

README.md
ResistantBear: Or "all" instead of "true". Of course we cannot exclude a folder named "true" or "all" then and I am not sure if such an option would be intuitive.

resources/c++/enums_and_unions.cpp
ResistantBear: Next example will be french or italian then? :D 
But I can change that of course

src/parser/helper/Helper.test.ts
ResistantBear: If it works, it would bypass the constructor of Configuration which does actually do more things than just assinging values. That would be really unelegant.

I have looked around a bit, and using an interface instead for the parameters is a choice if one wants to have named parameters.

src/app.ts
ResistantBear: Ok. Then for now, I will document that for both. We should remember to also change the parse-h-as-c option to also accept patterns once we change that for exclusions.

resources/c++/enums_and_unions.cpp
ResistantBear: Done

src/parser/helper/Language.ts
ResistantBear: I have talked with the others about it and they find `parseAllHAsC` acceptable. So I will go with that.

README.md
ce-bo: Hm okay, then let's keep it separated.

src/parser/helper/Helper.test.ts
ResistantBear: I have now introduced the interface `ConfigurationParams` for this.

test/metric-end-results/TestHelper.ts
ResistantBear: I am not sure if this works as expected with classes, especially with ones that have a non-standard constructor like `Configuration`. But I now have used this Partial<...>` utility type to merge multiple objects of the new `ConfigurationParams` interface mentioned above to achive the same goal.

src/app.ts
ResistantBear: Documentation for all command line options added to Readme

src/app.ts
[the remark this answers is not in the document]
ResistantBear: Added multiple warnings that `--parse-dependencies` is experimental, very slow and not ready for productive use yet.

src/parser/metrics/Metric.ts
ResistantBear: Changed.

src/parser/helper/Language.ts
ResistantBear: A "look in this map first and if there is no value, look in the second one" helper function, or what do you mean?

src/parser/metrics/Classes.ts
ResistantBear: Done

src/app.ts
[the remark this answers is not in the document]
ResistantBear: Added note about the experimental status to `--help` as well now.

README.md
```diff
@@ -67,7 +62,8 @@ Counts expressions that branch the control flow, like if-statements, loops, swit
 The number of function declarations inside a file. Includes all kinds of functions, like constructors, lambda functions, member functions, abstract functions, constructors, init blocks, closures, etc.
 
 **classes**<br>
-The number of class definitions inside a file, also counting for enums, interfaces, structs, traits, records and types.
+<<<<<<< HEAD
+The number of class definitions inside a file, also counting for enums, interfaces, structs, unions, traits and records.
```
ce-bo, line 66: Please fix this merge conflict statement

src/parser/Configuration.test.ts
```diff
@@ -1,33 +1,88 @@
...
+            expect(config.exclusions.size).toBe(0);
+
+            expect(config.parseAllHAsC).toBe(true);
+            expect(config.exclusions.size).toBe(0);
+            expect(config.parseSomeHAsC.size).toBe(0);
+        });
+
+        it("should work correctly when exclusions are given and some .h files should be parsed as C", () => {
+            const config = getTestConfiguration("sourcesPath", {
+                exclusions: "folder1 , folder2",
```
ce-bo, line 74: Nice!

src/parser/helper/Language.ts
ce-bo: In Line 139 there is an else-block which is executed on default, so at least the else-block should not be needed.

src/parser/helper/Language.ts
ce-bo: Well, what is this function actually doing if you would explain it to me? That could be the function name then.

ce-bo (approved):
Early LGTM, but please fix the readme before merging. :-)

src/parser/helper/Language.ts
ResistantBear: The question was more related to the scope of the function to extract here.
I am not sure if a function that first looks in one map and if there is no value checks another map with the key converted to lower case is something that we will ever use outside of this function. Converting to lower case and then looking up a value in one map, perhaps.

ResistantBear: I have now introduced a helper function `lookupLowerCase`. However, I am not sure if we ever can use that somewhere else and it does only contain two lines of code, so I am unsure if that is a good decision.

Issue #168, opened 2024-02-19
Extend documentation of metrics definitions (per language)


ce-bo:
As a Software Auditor, I want to exactly know which language expression has been counted for a metric, so that I can explain results in detail.

I suggest to add a list of supported language feature examples per metric. Similar to what Sonar is showing. The following examples makes clear what exactly is considered for a metric of a certain language:

## Java:
* Copmlexity: Supporting `for(var item : items)`, `for (int i = 0; i < 10; i++)`, etc; Not supported: ...
* Classes: Supporting `classes`, ....; Not Supported: `not_supported_type` (reason)

ResistantBear:
I am a bit confused here. If I understand correctly, this is distinct from #6 because it does not refer to the concrete expression in a code file, but to the a type of expression we count for in general?
So we should find out and document what expressions exactly do get parsed e.g. as "function_declaration" by tree-sitter if we count "function_declaration"? Could be a rather long list if there are so many ways to define a function, also requiring to look into the grammar to ensure that we have not missed to mention any kind of syntax to define a function.

ce-bo:
Correct, this is distsinct to #6. It refers to real code examples, that are supported. I added an example on how it should look like.
Let's discuss my idea. I think we as developers benefit from it as well as users. It makes it obvious which language features are already supported and which not. This leads to full transparency.

mylinhdao:
Splitted #202 for TypeScript and TSX

ResistantBear:
One file per language, linked in the readme

mylinhdao:
Splitted #210 to document TS/TSX

mylinhdao:
Splitted #243 to document Java

mylinhdao:
Splitted #262 for JavaScript

mylinhdao:
Splitterd #275 for Python

mylinhdao:
Splitted #277 for PHP

mylinhdao:
Splitted #295 for C#

clemens-mw:
Splitted #310 for Go

clemens-mw:
Splitted #311 for Kotlin

mylinhdao:
Splitted #326 for Bash, #327 for Ruby, #328 for Rust and #329 for JSON and YAML

mylinhdao:
I close this issue because all currently supported

Issue #88, opened 2024-02-07
May further redefine the Complexity metric


ResistantBear:
There are still points of disussion left, including:

- [ ] Whether to include throw-statements. Currently, we do no longer include them, like Sonar does for Java. Sonar does include them for JS/TS, however. So we cannot use Sonar as orientation here. Not counting for them is closer to the original MCC definition, as they do not branch the code, but unconditionally jump to somewhere, just like a `goto`.
- [ ] Whether to include return statements and goto statements. We currently do not do so and Sonar does seem to only include GOTO statements in a language where they can be conditional.
- [ ] Whether to include catch-blocks. As they can cause a branch of the code based upon the type of the thrown error, they probably can be counted when following the original MCC definition, but that whole throw/catch-thing is rather tricky when it comes to the control flow. 
- [ ] Whether or not to count the conditional C(++) preprocessor directives like `#if`, `#ifdef` or `#ifndef`, which are covered by the two syntax node types "preproc_if" and "preproc_ifdef"

ce-bo:
I put it to state "new issue" since we need to prepare/refine the issue.

ResistantBear:
Method references seem to be the Java-concept for passing a function by value. As that does not involve implementing a new function (at least not from the view of the programmer), we should not count them as new function definition.

_Originally posted by @ResistantBear in https://github.com/MaibornWolff/metric-gardener/issues/94#issuecomment-1934219476_

PR #199, opened 2024-03-05
Support parsing TSX code

### LINKED ISSUE #9: Support parsing TSX code
As described [here](https://github.com/tree-sitter/tree-sitter-typescript) tsx is a different dialect than typescript.
Therefore we need to support both grammars.

mylinhdao:
# Support parsing TSX code with tests
 Closes #9
 
 ## Description
 * Added tests to support TSX parser.
 * For **TSX** and **TypeScript** files, type declaration are counted for classes-metric if they have an implementation. Type alias will not be counted. 
_E.g_: Expressions like `type name = any` are not counted, but `type name = {firstname: string, lastname: string}` will be counted. 
 ## Definition of Done
 A PR is only ready for merge once all the following acceptance criteria are fulfilled:
 
 * [x]  All TODOs related to this PR have been closed
 * [x]  There are automated tests for newly written code and bug fixes
 * [x]  All bugs discovered while working on this PR have been submitted as issues (if not already an open issue)
 * [x]  Documentation ([README.md](https://github.com/MaibornWolff/metric-gardener/blob/main/README.md), [UPDATE_GRAMMARS.md](https://github.com/MaibornWolff/metric-gardener/blob/main/UPDATE_GRAMMARS.md)) has been updated


ce-bo:
Would you please add a short description what this PR is changing? I think multiple things are done here, not only adding tests if I am not mistaken, right?

mylinhdao:
@ce-bo yes, you are right. I have added the description.

README.md
```diff
@@ -67,7 +67,7 @@ Counts expressions that branch the control flow, like if-statements, loops, swit
 The number of function declarations inside a file. Includes all kinds of functions, like constructors, lambda functions, member functions, abstract functions, constructors, init blocks, closures, etc.
 
 **classes**<br>
-The number of class definitions inside a file, also counting for enums, interfaces, structs, traits and records.
+The number of class definitions inside a file, also counting for enums, interfaces, structs, traits, records and types.
```
ce-bo, line 70: I think we should add the information for TS that type aliases are not counted as you did in the PR description.

ce-bo: Could you please also check the README and mention there that TSX is supported now?

README.md
```diff
@@ -41,7 +41,7 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 -   Go
 -   PHP
--   TypeScript
+-   TypeScript, TSX (only type declaration using object type literal is counted for 'classes' metric, but not type alias.)
```
ce-bo, line 44: I think the description in the parenthesis would be redundant because #203 is merged already.

ce-bo (approved):

src/parser/metrics/Classes.ts
```diff
@@ -14,7 +17,27 @@ export class Classes implements Metric {
...
+        for (const nodeType of allNodeTypes) {
+            if (nodeType.metrics.includes(this.getName())) {
+                if (nodeType.expression === "type_alias_declaration") {
+                    this.statementsSuperSet.push(
+                        new SimpleLanguageSpecificQueryStatement(
+                            "(type_alias_declaration value: (object_type))",
+                            nodeType.languages,
+                            nodeType.activated_for_languages,
+                        ),
+                    );
```
ResistantBear, line 29: Do you expect that this specific `(type_alias_declaration value: (object_type))` can also occur in other languages? Otherwise, we might just use getQueryStatements(...), remove "classes" from "type_alias_declaration" in the nodeTypesConfig.json again and add the custom query for ts and tsx afterwards. I have done it that way for the custom queries for the C and C++ classes metric currently, we might want to do that consistently.

test/metric-end-results/TSXMetrics.test.ts
```diff
@@ -0,0 +1,49 @@
...
+            testFileMetric(tsxTestResourcesPath + "/class-likes.tsx", FileMetric.classes, 8);
+        });
+    });
+    describe("parses functions metric", () => {
+        it("should count all function declarations and lambda expressions.", () => {
+            testFileMetric(tsxTestResourcesPath + "/functions.tsx", FileMetric.functions, 8);
+        });
+    });
+    describe("parses complexity metric", () => {
+        it("should count all if, if-else, for, while, do-while labels.", () => {
```
ResistantBear, line 27: We only call case, default or goto jump labels "labels"

```suggestion
        it("should count all if, if-else, for, while and do-while statements.", () => {
```

test/metric-end-results/TSXMetrics.test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    beforeAll(async () => {
+        results = await parseAllFileMetrics(tsxTestResourcesPath);
+    });
+
+    describe("parses classes metric", () => {
+        it("should count all classes, interfaces, types, enum with implementation.", () => {
+            testFileMetric(tsxTestResourcesPath + "/class-likes.tsx", FileMetric.classes, 8);
+        });
+    });
```
ResistantBear, line 20: Line breaks for consistency :)

test/metric-end-results/TSXMetrics.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    describe("parses functions metric", () => {
+        it("should count all function declarations and lambda expressions.", () => {
+            testFileMetric(tsxTestResourcesPath + "/functions.tsx", FileMetric.functions, 8);
+        });
+    });
+    describe("parses complexity metric", () => {
+        it("should count all if, if-else, for, while, do-while labels.", () => {
+            testFileMetric(tsxTestResourcesPath + "/branches_loops.tsx", FileMetric.complexity, 9);
+        });
+        it("should count all ternary expression.", () => {
```
ResistantBear, line 30: Is this plural? Then it should be the following:
```suggestion
        it("should count all ternary expressions.", () => {
```

But I only see a single ternary expression in the tested file. If this is right, change it to " should count the ternary expression".

README.md
```diff
@@ -42,18 +42,20 @@ This is quite slow and can take up to one or two hours but can provide good resu
 -   Go
 -   PHP
 -   TypeScript
+-   TSX
```
ResistantBear, line 45: Should that really be in that category, or rather unter "Basic support"? Compared to the other languages here, we do not have that many tests for it. On the other hand, TSX is not dramatically different from TS, of course. Mh.

README.md
```diff
@@ -42,18 +42,20 @@ This is quite slow and can take up to one or two hours but can provide good resu
...
 -   Java
 -   C#
 -   C++
 -   Python
--   JavaScript
+-   JavaScript (without @flow annotation)
```
ResistantBear, line 50: I would keep it as "JavaScript" and add "Flow-annotated JavaScript" to the "Basic Support". Otherwise, we have to mention also other JavaScript dialects here that we might not support.

src/parser/metrics/Classes.ts
ce-bo: I thought about the same thing. I think, it is fine like this. With this approach it would be possible to reuse this logic for other languages that have a type_alias_decleration as well.

test/metric-end-results/TSXMetrics.test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    const testFileMetric = (inputPath, metric, expected) =>
+        expectFileMetric(results, inputPath, metric, expected);
+
+    beforeAll(async () => {
+        results = await parseAllFileMetrics(tsxTestResourcesPath);
+    });
+
+    describe("parses classes metric", () => {
+        it("should count all classes, interfaces, types, enum with implementation.", () => {
```
ResistantBear, line 17:
```suggestion
        it("should count all classes, interfaces, enums and object types with a type name.", () => {
```

README.md
ce-bo: I would like to have TSX as a stable language as mentioned in the Issue.

resources/tsx/class-likes.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type CustomValue = any;
+interface Props {
+  propA: CustomValue;
+}
+interface DefinedState {
+  otherStateField: string;
+}
+type State = DefinedState & ReturnType<typeof transformPropsToState>;
```
ResistantBear, line 36: Okay, whats that now?

src/parser/metrics/Classes.ts
ResistantBear: The benefit is that we can configure in the nodeTypesConfig for which languages we use this custom query. 
The disadvantage would be that we can then only treat `type_alias_declaration` with this custom query for ts and tsx now, which is unlikely to be applicable to any other language.

README.md
mylinhdao: Then i will add more tests to make it more stable

resources/tsx/class-likes.tsx
mylinhdao: that will not be counted :D but i will mention that in the doc at #168. 
But @ce-bo maybe you can decide explicitly if this should be counted :)

resources/tsx/functions.tsx
```diff
@@ -0,0 +1,31 @@
...
+        }}
+    />
+);
+const App = ({ message }: AppProps): React.JSX.Element => <div>{message}</div>; //lambda expression
+function useCallback<T extends (...args: any[]) => any>(
+    callback: T,
+    deps: []
+): T{
+    return useCallback(callback, deps);
+}
```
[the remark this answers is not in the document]
mylinhdao: ah my fault, forgot to rename that

resources/tsx/class-likes.tsx
ce-bo: This is a Union, right? So let's not count this way of defining an Alias .

ResistantBear: Is that really called a Union? We currently write in our Readme that we do count Unions as classes, because in C++, Unions can have member functions and so on. They are a strange memory saving thing otherwise, were you can have multiple variables that are all stored under the same memory location.

ResistantBear: Ok, these seem to be called `Intersection Types` in TS, so we should be fine.

src/parser/metrics/Classes.ts
```diff
@@ -17,27 +17,59 @@ export class Classes implements Metric {
...
-                            nodeType.expression,
-                            nodeType.languages,
-                            nodeType.activated_for_languages,
-                        ),
-                    );
-                }
-            }
-        }
+        this.statementsSuperSet = getQueryStatements(allNodeTypes, this.getName());
+        this.addQueriesForJS_TS_TSX();
```
ResistantBear, line 21: As @ce-bo did not want me to call the other function `addQueriesForC_C++`, this should be renamed, too.

```suggestion
        this.addQueriesForJsAndTsAndTsx();
```

src/parser/metrics/Classes.ts
```diff
@@ -58,7 +90,14 @@ export class Classes implements Metric {
...
     }
-
+    addQueriesForJS_TS_TSX() {
+        this.statementsSuperSet.push(
+            new SimpleLanguageSpecificQueryStatement(
+                "(type_alias_declaration value: (object_type))",
+                ["ts", "tsx"],
+            ),
+        );
+    }
```
ResistantBear, line 100: Please place the function directly below or above the similar `addQueriesForCAndCpp` function so that all special handling can be found easily.

resources/tsx/type.tsx
```diff
@@ -0,0 +1,13 @@
+import { string } from "yargs";
+type ObjectType = {
+    message: string;
+    name: string;
+    number: number;
+};
+type union = {a: 1} & {b: 2};
```
ResistantBear, line 7: Does that compile? Is `: 1` a type that represents the single number 1?

ResistantBear: Quite difficult. That are two object_types, too, I guess? But surrounded by whatever node is used for that intersection operator? I am not sure what would be the best approach to deal with that. But it is a quite exotic use case, probably.

resources/tsx/catch.tsx
```diff
@@ -0,0 +1,24 @@
+
+const parsePromises = new Set<Promise<string | null>>();
+
+parsePromises.add(
+    fetchData("www.google.com").catch((reason) => { //+0 complexity
```
ResistantBear, line 5: Good point that we do not count them as well currently. We should document that behaviour somewhere for the user, also that we do not count things like the build-in `forEach` function. 

For now, I have added that point to the list in issue #88 to may discuss later if we should change something about it.

test/metric-end-results/TSXMetrics.test.ts
```diff
@@ -14,36 +14,69 @@ describe("TSX metrics tests.", () => {
...
+            testFileMetric(tsxTestResourcesPath + "/branches_loops.tsx", FileMetric.complexity, 11);
         });
-        it("should count all ternary expression.", () => {
+        it("should count the ternary expression.", () => {
             testFileMetric(tsxTestResourcesPath + "/ternary.tsx", FileMetric.complexity, 2);
         });
+        it("should count the catch statement, but not the function call catch().", () => {
+            testFileMetric(tsxTestResourcesPath + "/catch.tsx", FileMetric.complexity, 6);
+        });
+        it("should also count the arrow function in react element.", () => {
```
ResistantBear, line 44:
```suggestion
        it("should also count the arrow function in the react element.", () => {
```

test/metric-end-results/TSXMetrics.test.ts
```diff
@@ -14,36 +14,69 @@ describe("TSX metrics tests.", () => {
...
+            testFileMetric(tsxTestResourcesPath + "/ternaryReact.tsx", FileMetric.complexity, 2);
+        });
     });
+
     describe("parses comment lines metric", () => {
         it("should count all multiline, one-line and inline comments.", () => {
-            testFileMetric(tsxTestResourcesPath + "/comments.tsx", FileMetric.commentLines, 6);
+            testFileMetric(tsxTestResourcesPath + "/comments.tsx", FileMetric.commentLines, 11);
+        });
+        it("should count no comment lines when file is empty.", () => {
```
ResistantBear, line 60:
```suggestion
        it("should count no comment lines when the file is empty.", () => {
```

test/metric-end-results/TSXMetrics.test.ts
```diff
@@ -14,36 +14,69 @@ describe("TSX metrics tests.", () => {
...
+            testFileMetric(tsxTestResourcesPath + "/empty.tsx", FileMetric.commentLines, 0);
         });
     });
+
     describe("parses real lines of code metric", () => {
         it("should not count comments and empty lines.", () => {
             testFileMetric(tsxTestResourcesPath + "/comments.tsx", FileMetric.realLinesOfCode, 16);
         });
+        it("should count no real lines of code when file is empty.", () => {
```
ResistantBear, line 69:
```suggestion
        it("should count no real lines of code when the file is empty.", () => {
```

test/metric-end-results/TSXMetrics.test.ts
```diff
@@ -14,36 +14,69 @@ describe("TSX metrics tests.", () => {
...
+            testFileMetric(tsxTestResourcesPath + "/empty.tsx", FileMetric.realLinesOfCode, 0);
+        });
     });
+
     describe("parses lines of code metric", () => {
         it("should count all lines in file.", () => {
-            testFileMetric(tsxTestResourcesPath + "/comments.tsx", FileMetric.linesOfCode, 22);
+            testFileMetric(tsxTestResourcesPath + "/comments.tsx", FileMetric.linesOfCode, 25);
+        });
+        it("should count 1 line of code when file is empty.", () => {
```
ResistantBear, line 78:
```suggestion
        it("should count 1 line of code when the file is empty.", () => {
```

resources/tsx/class-likes.tsx
ResistantBear: Ok, there are, of course, also the more commonly used `Union Types` via the `|` operator. Well, we have already decided to add language-specific documentation, so we should document that `Unions` are counted in C and C++, but the `Unions` in TS are something else and we do not count them, I guess? @ce-bo  @mylinhdao

resources/tsx/functions.tsx
```diff
@@ -10,15 +10,15 @@ const el = (
     />
 );
 const App = ({ message }: AppProps): React.JSX.Element => <div>{message}</div>; //lambda expression
-function useCallback<T extends (...args: any[]) => any>(
+function useCallbackCustom<T extends (...args: any[]) => any>(
     callback: T,
     deps: []
 ): T{
     return useCallback(callback, deps);
```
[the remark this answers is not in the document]
mylinhdao: no, `useCallback` is from the react library. I forgot to import it.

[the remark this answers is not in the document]
ResistantBear: Ah, ok, fine

resources/tsx/type.tsx
ResistantBear: Further discussion about this topic is now under issue #213

src/parser/metrics/Classes.ts
ResistantBear: We now settled on not doing this via the nodeTypesConfig. It would be really intransparent to "add" a node here to a metric for a language but then some special handling kind of "catches" that and does something unexpected that was programmed to work for one language only. For the unlikely case that someone wants to add another language to be handled by this exact same special logic, it can be added in the source code.

resources/tsx/class-likes.tsx
ResistantBear: Further discussion about this topic is now under #213

ResistantBear (approved):
All remarks resolved

src/parser/metrics/Classes.ts
```diff
@@ -71,7 +71,14 @@ export class Classes implements Metric {
             ),
         );
     }
-
+    addQueriesForTSAndTSX() {
```
ResistantBear, line 74:
```suggestion
    
    addQueriesForTSAndTSX() {
```

Issue #210, opened 2024-03-12
Add documentation file for TS/TSX


mylinhdao:
Implement issue #168 for TypeScript/TSX.

PR #226, opened 2024-03-15
Add documentation file for TS/TSX

mylinhdao:
# Add documentation file for TS/TSX
Closes: #210 

## Description
Add an file, which documents what we exactly count for each file metric. This file is linked in the README.md.

## Definition of Done

A PR is only ready for merge once all the following acceptance criteria are fulfilled:
* [x]  All TODOs related to this PR have been closed
* [x]  There are automated tests for newly written code and bug fixes
* [x]  All bugs discovered while working on this PR have been submitted as issues (if not already an open issue)
* [x]  Documentation ([README.md](https://github.com/MaibornWolff/metric-gardener/blob/main/README.md), [UPDATE_GRAMMARS.md](https://github.com/MaibornWolff/metric-gardener/blob/main/UPDATE_GRAMMARS.md)) has been updated


TS_TSX.md
```diff
@@ -0,0 +1,48 @@
+# File Metrics for TypeScript (TS) and TypeScript JSX (TSX)
+
+This document outlines the definition of file metrics for TypeScript (TS) and TypeScript JSX (TSX).
+
+### 1. classes
+
+The "classes" metric counts:
+
+-   keywords like: `class`, `abstract class`, `interface`, `enum`
```
ResistantBear, line 9: Is it correct to say that we really count the keyword? Technically, we do not, as we count the nodes `class_declaration` and not the (unnamed) node that contains the keyword itself. 
Practically, I don't know if the keywords `class`, etc. can be only used to define a class in TS or also on other places, so that that your documentation is right anyway or not. For C and C++, this would be clearly wrong, as we can do things like 
````C++
#include <iostream>

class class_name;

class class_name {
    public:
        void print_hello() {
            std::cout << "Hello!" << std::endl;
        }
};

int main() {
    class class_name* object = new class_name();
    class class_name object2();
    class class_name object3();

    object->print_hello();
    object2.print_hello();
    object3.print_hello();
}
````

Here, we actually do only count the thing with the second class keyword.

Comment (ResistantBear) [reply TS_TSX.md:9]: Also, what does "keywords like" mean? Do we also count other keywords that are similar? Then we should mention them here.

I would not write here that we count keywords at all, however. We should write that we count class declarations, abstract class declarations, interface declarations and enum declarations.

File: TS_TSX.md [hunk 2]
```diff
@@ -0,0 +1,48 @@
...
+-   type aliases with object type literals e.g. `type ObjectType = { message: string};`
```
ResistantBear, line 10: As we currently do not count them when there is more than one object type literal:
```suggestion
-   type aliases with a single object type literal, e.g. `type ObjectType = { message: string};`
```

TS_TSX.md
```diff
@@ -0,0 +1,48 @@
...
+    -   `type Union = {a: 1} & {b: 2};`
+    -   `type Intersect = {a: 1} | {b: 2};`
+    -   `type Tuple = [name: string, age: number];`
+    -   `type AliasType = number;`
+
+### 2. functions
+
+The "functions" metric counts:
+
+-   methods declarations in classes, abstract classes and interfaces
```
ResistantBear, line 24:
```suggestion
-   method declarations in classes, abstract classes and interfaces
```

TS_TSX.md
```diff
@@ -0,0 +1,48 @@
...
+-   static methods, constructors, getters and setters
```
ResistantBear, line 25: Whats with the static initialization block you mentioned recently?

TS_TSX.md
```diff
@@ -0,0 +1,48 @@
...
+-   functions declared with the keyword `function`
+-   arrow functions
+
+### 3. complexity
+
+The "complexity" metric counts:
+
+-   `if`, `if else` statements, but **no** `else` statements
```
ResistantBear, line 33: I don't know if we should explicitely mention that we count no `else` here, as that is already clear when reading the metric definition in the readme and when mentioning that we count for `if` and `if else`.

TS_TSX.md
```diff
@@ -0,0 +1,48 @@
...
+-   ternary expressions
+-   for, while, do-while loops
```
ResistantBear, line 35: We should write `do..while` just like they do here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/do...while?retiredLocale=de

Also a nit:  using "and" instead of a comma before the last element of the list is a better writing style.
```suggestion
-   `for`, `while` and `do...while` loops
```

ResistantBear: We should probably also mention `for...of`, `for await...of` and `for...in` loops here.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

TS_TSX.md
ResistantBear: The same holds for `default` and `finally` below

TS_TSX.md
```diff
@@ -0,0 +1,48 @@
...
+-   `case` labels in switch-statements, but **no** `default` labels
+-   `catch` labels in try-catch blocks, but **no** `finally` labels
+-   binary operations `&&` and `||`
```
ResistantBear, line 38: Like in the readme, we should make clear that we only count **logical** binary operations, not things like `+`, `-`, `*`, `/`, etc.
```suggestion
-   logical binary operations `&&` and `||`
```

TS_TSX.md
```diff
@@ -0,0 +1,48 @@
...
+-   everything counted for "functions" metric
```
ResistantBear, line 39:
```suggestion
-   everything counted for the "functions" metric
```

TS_TSX.md
```diff
@@ -0,0 +1,48 @@
...
+
+It does **not** count:
+
+-   `.catch()` method call for Promises
+-   `.forEach()` method call for iterables
```
ResistantBear, line 44: There are also similar functions like `map`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map. We should make clear that do not count them as well:
```suggestion
-   `.forEach()` method call for iterables and similar build-in functions
```

README.md
```diff
@@ -35,7 +35,7 @@ This is quite slow and can take up to one or two hours but can provide good resu
 
 -   Go
 -   PHP
--   TypeScript
+-   [TypeScript and TSX](https://github.com/MaibornWolff/metric-gardener/blob/main/TS_TSX.md)
```
ResistantBear, line 38: Maybe we should continue to list them separately and link them to the same document? But I'm not sure, as both are quite similar.

ResistantBear: Also, I am not sure if it is that clear that we can click on the language name to find information about what we count for that language. We maybe should add somewhere in the readme a note that one can find language-specific details about the metrics in these files.

TS_TSX.md
```diff
@@ -0,0 +1,48 @@
+# File Metrics for TypeScript (TS) and TypeScript JSX (TSX)
+
+This document outlines the definition of file metrics for TypeScript (TS) and TypeScript JSX (TSX).
+
+### 1. classes
```
ResistantBear, line 5: We do not have numbers for them in the main Readme. So I think, we should not do this here, either. Also, the order of the metrics here does not match that of the readme. We should probably order them identical to the readme here.

ResistantBear (commented):
Please take care of the remarks.

TS_TSX.md
```diff
@@ -2,47 +2,47 @@
...
-The "classes" metric counts:
+The "complexity" metric counts:
 
--   keywords like: `class`, `abstract class`, `interface`, `enum`
--   type aliases with object type literals e.g. `type ObjectType = { message: string};`
+-   `if` and `if else` statements
+-   ternary expressions
+-   `for...of`, `for await...of`, `for...in`,`while` and `do...while` loops
+-   `case` labels in switch-statements
+-   `catch` labels in try-catch blocks
```
ResistantBear, line 13: They are no label inside a try, but a block themself.
```suggestion
-   `catch` blocks
```

TS_TSX.md
```diff
@@ -2,47 +2,47 @@
...
+The "classes" metric counts:
 
--   `if`, `if else` statements, but **no** `else` statements
--   ternary expressions
--   for, while, do-while loops
--   `case` labels in switch-statements, but **no** `default` labels
--   `catch` labels in try-catch blocks, but **no** `finally` labels
--   binary operations `&&` and `||`
--   everything counted for "functions" metric
+-   declarations of class, abstract class, interface and enum
```
ResistantBear, line 35: Should be in plural
```suggestion
-   declarations of classes, abstract classes, interfaces and enums
```

TS_TSX.md
```diff
@@ -2,47 +2,47 @@
...
+-   other type aliases like
+    -   `type Union = {a: 1} & {b: 2};`
+    -   `type Intersect = {a: 1} | {b: 2};`
+    -   `type Tuple = [name: string, age: number];`
+    -   `type AliasType = number;`
 
-### 4. lines_of_code and real_lines_of_code
+### lines_of_code, comment_lines and real_lines_of_code
 
 see README.md
```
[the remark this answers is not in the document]
ResistantBear: That would be more cool

Issue #227, opened 2024-03-15
Make pre-commit check fail if there are lint warnings


clemens-mw:
Avoid having lint issues in pull requests

PR #259, opened 2024-03-20
Pre-Commit hook checks for lint warnings

clemens-mw:
# Pre-Commit hook checks for lint warnings

Closes: #227


mylinhdao:
Oddly, ` --fix` seems to fix nothing even when the documentation says it should fix linting issues automatically. So we think that it's ok to remove it.

mylinhdao (approved):

Issue #279, opened 2024-03-25
Use Node.js built-in zlib


clemens-mw:
https://nodejs.org/docs/latest-v18.x/api/zlib.html

PR #292, opened 2024-03-28
chore: remove unnecessary zlib #279

Arrekusanda:
# remove unnecessary zlib

Closes: #279

## Description
Node.js zlib is used


mylinhdao (approved):
You are right, we have used zlib in @types/node to compress output files. So we don't need zlib in package.json :)

PR #294, opened 2024-03-28
Fix wrong file ending when compressing

clemens-mw:
# Fix wrong file ending when compressing

Closes: #293

## Description

Fixed outputting the file name without .gz when printing the last output to the console.


Arrekusanda (approved):
lgtm

Issue #63, opened 2024-02-01
Possible improvement for the performance of real lines of code


ResistantBear:
I believe we could change the set to a counter and use the last line as memory. We increase the line count as long as they node's start position is higher than the current last entry. That would reduce the memory overhead.

_Originally posted by @BridgeAR in https://github.com/MaibornWolff/metric-gardener/pull/59#pullrequestreview-1854081254_

ResistantBear:
There is the branch "tech/63/optimize-rloc" with a not-yet-working implementation linked to this issue if someone wants to pick this up.

PR #314, opened 2024-04-10
Use Counter in Real Lines of Code to Improve Performance

clemens-mw:
# Use Counter in Real Lines of Code to Improve Performance

Closes: #63


mylinhdao (approved):
LGMT :)

